PR #12, opened 2015-01-13
Updated the navigation article.

hamorphis:


enchev:
:+1:

navigation.md
```diff
@@ -124,6 +124,25 @@ var navigationEntry = {
 };
 topmost.navigate(navigationEntry);
 ```
+#Navigating to another page and passing context
+Sometimes, the page being navigated to would have to receive information about the context in which this navigation happened. The bese example would be a master-details scenario where there are two pages -- the main page containing a list of some entities and a details page which provides details about a particular entity. In this case, when navigating to the details page it is mandatory to transfer some primary key or ID information about the entity the details page should show. This is done with the help of the **context** property of a NavigationEntry:
```
hshristov, line 128: bese - probably base

hshristov:
We need to update help with onNavigatedTo method.

PR #36, opened 2015-02-10
Ensure index html

ErjanGavalji:
Have the index page of index.md. Have a link, pointing to it, named README.md.

Now, if a content article needs to point to the root page, it'd need to only point to '/'


nsndeck:
:+1:

PR #39, opened 2015-02-12
Getting started revisions

tjvantoll:
Here is my suggested version of the getting started page. The rendered page can be viewed [here](https://github.com/tjvantoll/docs/blob/2bc959559496b189a91f7a6ba2ffef6422801725/getting-started.md) and is a lot easier to read through than this Markdown diff.

A couple of things:
- @valentinstoychev
  - Can you confirm that I'm using the correct syntax to get the JavaScript/TypeScript tabs that the rest of the documentation uses? I copied and pasted from another page, but I don't know if I need to do something to activate that behavior for the live site.
  - For me, the message label never gets updated on iOS (before I made any changes to this guide). It works fine in Android. I'm guessing this is something that's resolved with the beta?
- @burkeholland / @codylindley / everyone else that gives this a read
  - Please read through this and let me know what you think. Be harsh.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -1,87 +1,146 @@
 ---
-nav-title: "Getting Started With NativeScript"
-title: "Getting Started With NativeScript"
+nav-title: "Getting Started with NativeScript"
+title: "Getting Started with NativeScript"
 description: "NativeScript Documentation: Getting Started"
 position: 2
 ---
 
-# Getting Started With NativeScript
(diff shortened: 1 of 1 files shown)


tjvantoll:
I made a proofreading pass, incorporated @codylindley's suggestions, and verified this all works with 0.9. I'm just waiting on the team to verify that `tns create` will generate the structure that this guide expects in 0.9. After that this should be good to merge in.

If anyone else has feedback on this let me know.

tjvantoll:
Replaced by #50.

PR #52, opened 2015-03-03
Updated getting started article for the CLI

tjvantoll:
ping @ikoevska.

This has all of your suggested changes other than the TypeScript code. I agree it should be added but I can't get TypeScript to work at the moment. When I get it up and running I'll add it in.

Let me know if you think anything else should change here, and thanks again.


ikoevska:
Closing this one as we actually merged the content.

PR #56, opened 2015-03-05
A date that should not be

jamlfy:


vakrilov:
:+1: 
It shall be undone

PR #63, opened 2015-03-10
Adding a FAQ with common questions and answers

tjvantoll:


faq.md
```diff
@@ -0,0 +1,41 @@
+---
+nav-title: FAQ
+title: "FAQ"
+description: A list of frequently asked NativeScript questions and answers
+position: 19
```
ikoevska, line 5: Why not put this somewhere in the top resources? You can group it with the overview or the Getting Started?

faq.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+# NativeScript FAQ
+
+### What Android API level does NativeScript target?
+
+Currently the Android runtime is built against Android API level 17; therefore APIs from higher API levels are not currently supported through JavaScript, however, the NativeScript team is [updating the Android runtime to API level 21](https://github.com/NativeScript/android-runtime/issues/39) soon.
```
ikoevska: You can drop a few currently and however here.

The Android runtime is built against Android API level 17; therefore, APIs from higher API levelrs are not supported through JavaScript. The NativeScript team...

faq.md
```diff
@@ -0,0 +1,41 @@
...
+
+### How do I add a navigation bar to my iOS app?
+
+A NavBar element is planned for NativeScript 1.0. In the meantime you can use iOS-specific code to add a navigation bar to your app. For example, the following code adds a custom navigation bar with a title:
```
ikoevska, line 16: Just a general comment about such code blocks. Until NativeScript becomes mature and well-known, we need to throw in a few lines here and there explaining which code should be put where and what it does. My suggestion is to add comments in the code blocks which explain exactly what's happening.

faq.md
tjvantoll, line 5: To me this is reference material kind of like the troubleshooting article, and not something we would want people to see at the beginning. I think [this sort of FAQ](http://www.telerik.com/nativescript/faq) could be in the beginning, but these questions are very specific.

faq.md
tjvantoll: Thanks. Updated.

faq.md
tjvantoll, line 16: Good suggestion. I added a few comments to the code blocks. Let me know what you think.

faq.md
ikoevska, line 5: Yeah, I made this comment before actually going through the FAQ. My expectation was that this was a more "general" sort of FAQ =). Disregard this one.

faq.md
```diff
@@ -0,0 +1,47 @@
...
+
+    // Make sure we're on iOS before configuring the navigation bar
+    if (applicationModule.ios) {
+        page.ios.title = "My App";
+
+        var controller = frameModule.topmost().ios.controller;
+        var navBar = controller.navigationBar;
+
+        navBar.tintColor = new colorModule.Color("#FFFF00").ios;
+        navBar.barStyle = UIBarStyle.UIBarStyleBlack;
```
ikoevska, line 45: This seems like native stuff to me (I might be mistaken though, if so - disregard this).

If this is indeed native code, we might want to state that explicitly (in a comment, for example). The reason why I'm suggesting this is that we might have hybrid or web devs who don't known native code and native devs look at that. In any case, our dev users need to know they that can mix up native stuff in there to achieve gorgeous native-like UI :)

If this is not really native code, you might want to put in there some comment in the lines of that NativeScript provides very intuitive platform-specific abstractions so that if you're familiar with the native code, you will feel pretty comfortable. (You might put that in the FAQ area and instead emphasize that this is the native-like code mentioned above)

tjvantoll, line 45: Yep this is iOS specific. I added a few more comments. Do you think that helps?

ikoevska, line 45: Awesome! :) You have a :+1: for this PR from me.

atanasovg:
:+1:

PR #69, opened 2015-03-24
http POST JSON example

Andrew8xx8:


ErjanGavalji:
@Andrew8xx8, thanks for the addition!
The NativeScript Api Reference as well as the code examples get generated automatically, hence we won't directly merge your pull request. We applied it via the https://github.com/NativeScript/NativeScript/pull/90 and https://github.com/NativeScript/NativeScript/pull/93 though

PR #70, opened 2015-03-24
Add SET for creating an environment variable

gabesumner:
SET is required to create a new environment variable on Windows

### DIFF
--- a/setup/ns-cli-setup/ns-setup-win.md
+++ b/setup/ns-cli-setup/ns-setup-win.md
@@ -50,7 +50,7 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
         JAVA_HOME=Path to the jdk* install folder
         ```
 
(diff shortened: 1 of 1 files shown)


setup/ns-cli-setup/ns-setup-win.md
```diff
@@ -50,7 +50,7 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
         JAVA_HOME=Path to the jdk* install folder
```
ikoevska, line 50: SET should be added here as well

Shell should be added in the opening syntax for the code block.

ErjanGavalji:
Closing the pull request. The command must actually be SETX to be stored permanently. Additionally, to have it working for all users, the /M switch must be passed. For the /M switch to work, the console must be started as administrator.

Changes are applied.
Thanks for the notice!

PR #80, opened 2015-04-21
Removed mentions of nodejs-legacy

ikoevska:
Replaced them with links to official Node install documentation for Linux


ikoevska:
ping @ErjanGavalji

PR #88, opened 2015-05-12
Update "Receive continuous location updates"

shenningsgard:
"new LocationManager()" returns an error


ErjanGavalji:
@shenningsgard Thanks for the notice!

The ApiReference folder of the documentation gets generated automatically from the [NativeScript](https://github.com/nativescript/nativescript) repo with the HowTo article snippets extracted from files in the apps/tests folder.

That said, I fixed the require statements in the original location. They will get applied live with the next update of the ApiReference.

I'm closing the issue now.

Cheers,
Erjan Gavalji

PR #102, opened 2015-06-10
Introducing NativeScript plugins

ikoevska:
Please, merge this after https://github.com/NativeScript/nativescript-cli/pull/547 gets merged. We need to wait for a fix from Fatme so that the merged xml example is valid.


ErjanGavalji:
Article now retrieved automatically from the nativescript-cli/ repo. Closing the issue.

PR #131, opened 2015-08-27
Update navigation docs

dmccuskey:
i made some modifications to this page. please review.


navigation.md
```diff
@@ -121,11 +121,11 @@ To navigate between pages, you can use the [`navigate`](ApiReference/ui/frame/RE
 
 In addition, each `Page` instance carries information about the frame object which navigated to it in the `frame` property. This lets you navigate with the `frame` property as well. 
 
+
 ### The Topmost Frame
 
-The topmost frame is the frame at the root level. When navigating inside your app, you can do it with the topmost frame.
+The topmost frame is the root-level container for your app's UI and it can be used to navigate inside of your app. You can get a reference to this frame by using the `topmost()` method of the Frame module.
```
ikoevska: and you can use it to navigate inside your app

navigation.md
```diff
@@ -147,9 +150,9 @@ topmost.navigate("details-page");
 topmost.navigate("details-page");
 ```
 
-### Navigate with Factory Function
+### Navigate Using a Function
 
-You can specify the page to which you want to navigate by a factory function which returns the page instance.
+A more dynamic way of navigating can be done by providing a function that returns the instance of the page to which you want to navigate.
```
ikoevska, line 157: A more dynamic way of navigating is by providing a

navigation.md
dmccuskey: updated.

question: is the top frame the ONLY way to navigate within an app when using NS UI structure ?

navigation.md
dmccuskey, line 157: i don't think your edit makes sense. here are some examples which have a simpler subject to help show why:

Cleaning can be done by providing a broom
Cleaning is by providing a broom

An orange can be cut by using a knife 
An orange is by using a knife

enchev:
:+1:

ErjanGavalji:
Both are applicable.
:+1:

PR #128, opened 2015-08-19
New quick start

tjvantoll:
Ping @atanasovg @hdeshev 

I was able to get the Jekyll setup working which helped a lot for testing this. Here's the side nav I was talking about:

screen shot 2015-08-19 at 1 09 19 pm

The two problems are that, 1) this requires a bit of CSS and JavaScript to get working (see the top and bottom of [this file](https://raw.githubusercontent.com/tjvantoll/docs/83e2ac5899c4035f4c8c89e4f8eb5a8fe08eadad/getting-started.md)), and 2) the nav isn't fixed like it is on http://nativescript.github.io/quick-start/. I can make it work, but it gets really hacky as I basically have to reimplement the site's layout for this page.

The being said, other than those two things I was pleasantly surprised by how easy it was to port this content and get everything working. Let me know what you think.

PR #134, opened 2015-09-01
Split quick start

tjvantoll:
Ping @ikoevska @jlooper 

I split things up for sanity, and put this in a new PR so we can actually open this. Are there any outstanding issues from https://github.com/NativeScript/docs/pull/128 that we need to address?


tjvantoll:
I just verified that the include syntax is working as intended with some big help from @hdeshev. Once @ikoevska signs off on the content we should be good to go.

getting-started/chapter-0.md
```diff
@@ -0,0 +1,26 @@
+## NativeScript Getting Started Guide
+
+Welcome to the [NativeScript](https://nativescript.org) getting started guide. In this guided tutorial you'll use NativeScript, a cross-platform JavaScript framework for building native mobile apps, to build an iOS and Android app from scratch.
```
ikoevska: 1st sentence -> getting started guide
2nd sentence -> guided tutorial

You can just drop the guided part (as it is pretty obvious :))

getting-started/chapter-0.md
```diff
@@ -0,0 +1,26 @@
...
+
+> **TIP:** For a video introduction to NativeScript, see the [getting started guide on YouTube](https://www.youtube.com/watch?v=rsCT5fpES4Q).
+
+### What you're building
+
+This guide will walk you through building [Groceries](https://github.com/NativeScript/sample-Groceries), a groceries management app that does the following things:
+
+- Connects to an existing RESTful service.
+- Lets users to register and login.
```
ikoevska: Lets users register and login. (alternatively, "Provides user registration and login.")

getting-started/chapter-0.md
```diff
@@ -0,0 +1,26 @@
...
+- Lets authenticated users to add and delete groceries from a list.
```
ikoevska: Lets authenticated users add and delete groceries. (please, look through instances of allow -> let replacements and remove the unnecessary "to" leftovers :))

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Install NativeScript and configure your environment
+
+The NativeScript CLI has a few system requirements you must have in place before building NativeScript apps. As a first step, start by going through the instructions for your operating system:
+
+- [Windows](http://docs.nativescript.org/setup/ns-cli-setup/ns-setup-win.html)
+- [OS X](http://docs.nativescript.org/setup/ns-cli-setup/ns-setup-os-x.html)
+- [Linux](http://docs.nativescript.org/setup/ns-cli-setup/ns-setup-linux.html)
+
+> **TIP**: If you're a bit overwhelmed by these requirements, or if you're looking for a way to build iOS apps on Windows, you might be interested in [using NativeScript with Telerik AppBuilder](http://docs.nativescript.org/setup/quick-setup#the-appbuilder-tool-set). Telerik AppBuilder provides tooling for NativeScript apps, including the ability to perform iOS and Android builds in the cloud, which removes the need to complete these system requirements.
```
ikoevska: This is probably a marketing question with Project Zero coming along and all... Should we replace "Telerik AppBuilder" with "Telerik Platform"?

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+After completing the setup you should have two commands available from your terminal: `tns`—which is short for **T**elerik **N**ative**S**cript—and `nativescript`. The two commands are equivalent, so we'll stick with the shorter `tns`.
```
ikoevska, line 15: ОК, the diff has some issues processing the bold formatting but it looks ok rendered.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+git clone https://github.com/NativeScript/sample-Groceries.git
+```
+
+After that, change to the newly cloned repo's folder:
+
+```
+cd sample-Groceries
+```
+
+The master branch has the final state of the Groceries app. Feel free to [refer back to it](https://github.com/NativeScript/sample-Groceries) at any time, but for now, switch over to the “start” branch for this guide's starting point:
```
ikoevska, line 57: I haven't gotten to the last chapter but if it's not already there, you should provide this reference there as well. By the last chapter, users will have forgotten that they can look at the final state in the master branch of the repo.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+```
```
ikoevska, line 59: I suggest moving the git checkout next to the git clone and then switching to the working directory.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+Next, add the Android platform with the same `platform add` command:
+
+```
+tns platform add android
+```
+
+<div class="exercise-end"></div>
+
+>**NOTE:** You can add platforms only for SDKs that you already have installed on your development machine. If you get errors running `tns platform add`, refer back to the section on [setting up your development environment](#install-nativescript-and-configure-your-environment).
```
ikoevska: This is more of an IMPORTANT note in my opinion as it is rather significant.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+The `platform add` command adds a folder called `platforms` to your project, and copies all of the required native SDKs into this folder. When you build the application, the NativeScript CLI will copy your application code into the `platforms` folder so that a native binary can be created.
+
+### Running your app
+
+With the platform initialization complete, you can run your app in an emulator or on devices.
```
ikoevska, line 93: completes

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+![iOS login](img/cli-getting-started/chapter1/ios/1.png)
+
+Next, run your app on an Android emulator with the following command:
+
+```
+tns run android --emulator
+```
+
+> **WARNING**:
```
ikoevska, line 115: Double-check if our Jekyll processes this correctly. If it doesn't render the un-numbered list properly, try reworking it with html tags instead.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+![Android login](img/cli-getting-started/chapter1/android/1.png)
+
+<div class="exercise-end"></div>
+
+Here are a few other tips for running NativeScript apps.
+
+> **TIP**:
+> * To run on a USB-connected Android or iOS device, use the same `run` command without the `--emulator` flag—i.e. `tns run android` and `tns run ios`.
+> * The `tns device` command lists all USB-connected iOS devices, USB-connected Android devices, and Genymotion virtual devices that `tns run` can deploy to. Note that `tns device` does not list iOS simulators or Android AVDs.
```
ikoevska: The team says it does list running Android AVDs :)

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Development workflow
+
+At this point, you have the NativeScript CLI downloaded and installed, as well as the iOS and Android dependencies that you need to run your app. Now you need a good workflow that lets you make changes and see results fast.
+
+The simplest way to see your changes is to execute `tns run ios` or `tns run android` after you save files. To see this action, let's make a trivial update to your app.
```
ikoevska: This is not really the simplest way. LiveSync is.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+If your previous `tns run ios` or `tns run android` task is still running, type `Ctrl+C` in your terminal to kill it.
+
+Open your app's `app/views/login/login.xml` file in your text editor of choice and change `<Label text="hello world" />` to `<Label text="hello NativeScript" />`.
+
+Return to your terminal and run either `tns run ios --emulator` (if you're on a Mac), or `tns run android --emulator`. You should see the app relaunch and the updated text displayed.
+
+<div class="exercise-end"></div>
+
+As you might have noticed, the `tns run` command blocks your terminal while your app is running (it's the task you had to use `Ctrl` + `C` to kill). The `tns run` command sticks around because the task is a console of sorts. The task shows both the output of `console.log()` statements as your app executes, as well as stack traces when things go wrong. So if your app crashes at any time during this guide, look to the terminal for a detailed report of the problem.
```
ikoevska: I think you can drop the "The `tns run` command sticks around because the task is a console of sorts." part.

getting-started/chapter-1.md
```diff
@@ -0,0 +1,173 @@
...
+
+The iOS and Android logs can be a bit noisy, so you might have to scroll up a bit to find the actual problem. For example if I try to call `foo.bar()` when `foo` does not exist, here's the pertinent information I get on iOS:
```
ikoevska: Do you really need "pertinent" here? You can go with just information here.

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+```
+.
+└── sample-Groceries
+    ├── app
+    │   └── ...
+    ├── package.json
+    └── platforms
+        ├── android
+        └── ios
```
ikoevska, line 19: Users on Windows will not have the ios folder. I am not sure whether we should state this.

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+``` JavaScript
+var applicationModule = require("application");
+applicationModule.mainModule = "./views/login/login";
+applicationModule.start();
+```
+
+Here, you're requiring, or importing, the [NativeScript application module](http://docs.nativescript.org/ApiReference/application/HOW-TO). Then, you set its `mainModule`, or the starting screen of your app to be the login screen, which lives in your app's `views/login` folder.
+
+> **TIP**: JavaScript modules in NativeScript follow the [CommonJS specification](http://wiki.commonjs.org/wiki/CommonJS). This means you can use the [`require()` method](http://wiki.commonjs.org/wiki/Modules/1.1#Module_Context) to import modules, as is done above, as well as use the `export` keyword to expose a module's properties and methods, which we'll look at later in this chapter. These are the same constructs Node.js uses for JavaScript modules, so if you already know how to use Node.js modules, you already know how to use NativeScript modules.
```
ikoevska: In the last sentence there are two already-s. Remove one of them (probably removing the first works best).

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+Let's dig into the files used to create your app's UI, which reside in the `app/views` folder. Each folder in `app/views` contains the code for one of the three pages in Groceries: `list`, `login`, and `register`. If you look in the `app/views/login` folder, you'll see three files: `login.css`, `login.js`, and the `login.xml` file we updated in the previous chapter. When you open `login.xml` again you should see the following code:
+
+``` XML
+<Page>
+    <Label text="hello NativeScript" />
+</Page>
+```
+
+This page currently contains two UI components: a `<Page>` and a `<Label>`. To make this page look more like a login page, let's add a few additional components, namely two `<TextField>` elements and two `<Button>` elements.
```
ikoevska, line 85: I'm wondering if saying "namely username and password fields and login and register buttons" instead of the exact names of the UI components?

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+<h4 class="exercise-start">
+    <b>Exercise</b>: Add UI components to <code>login.xml</code>
+</h4>
+
+Open `app/views/login/login.xml` and replace the existing `<Label>` with the following code:
+
+``` XML
+<TextField hint="Email Address" keyboardType="email" />
+<TextField secure="true" hint="Password" />
```
ikoevska: hm, I'd move hint before secure for consistency with the syntax above.

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+- The [Stack Layout](https://docs.nativescript.org/ApiReference/ui/layouts/stack-layout/HOW-TO.html) lets you stack child UI components either vertically or horizontally.
+- The [Wrap Layout](https://docs.nativescript.org/ApiReference/ui/layouts/wrap-layout/HOW-TO.html) allows child UI components to flow from one row or column to the next when space is filled.
+
+For your login screen, all you need is a simple `<StackLayout>` to stack the UI components on top of each other. In later sections, you'll use some of the more advanced layouts.
+
+<h4 class="exercise-start">
+    <b>Exercise</b>: Add a stack layout to the login screen
+</h4>
+
+In `login.xml`, add the `<StackLayout>` component below directly within the `<Page>` component. `login.xml` should look like this:
```
ikoevska: This "below directly within" sounds messy and unclear. :) Please, rephrase.

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+``` XML
+<Page>
+    <StackLayout orientation="vertical">
+
+        <TextField hint="Email Address" keyboardType="email" />
+        <TextField secure="true" hint="Password" />
```
ikoevska: If you decide to apply my comment about switching the places of secure and hint, make sure to do it here as well. =)

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+<div class="exercise-end"></div>
+
+The stack layout is a UI component, and as such, it has attributes just like the `<TextField>` and `<Button>` components you used in the previous section. Here, the `orientation="vertical"` attribute tells the stack layout to arrange its child components vertically.
+
+After you run your app with this change, you'll see that your login page's UI components stack up:
+
+![login 2](img/cli-getting-started/chapter2/ios/2.png)
+![login 2](img/cli-getting-started/chapter2/android/2.png)
+
+Although the UI components are in the correct order, they could use some spacing and a bit of color to make the app look a bit nicer. To do that let's look at another NativeScript feature: CSS.
```
ikoevska: Double "a bit" in one sentence. Probably go with "and some color to make the app look nicer"

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+> **TIP**:
```
ikoevska: I'd move this tip closer to the list of layouts, as it's more relevant there.

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+> * Refer to the NativeScript docs for a [more detailed look at how NativeScript layouts work](http://docs.nativescript.org/layouts) and the various things you can do to configure them.
+
+### CSS
+
+NativeScript uses a [subset of CSS](http://docs.nativescript.org/styling) to change the visual appearance of your app. You can use three mechanisms to add CSS properties to UI components: [application-wide CSS](http://docs.nativescript.org/styling#application-wide-css) (`app.css`), [page-specific CSS](http://docs.nativescript.org/styling#page-specific-css), and an [inline `style` attribute](http://docs.nativescript.org/styling#inline-css).
+
+> **TIP**:
+> * Place CSS rules that should apply to all pages in your `app.css`, and CSS rules that apply to a single page in a page-specific CSS file (e.g. `login.css`). 
+> * Although inline styles are great for quick testing—e.g. `<Page style="background-color: green;">` you should avoid them in general because the `style` attributes tend to clutter up XML files especially if/when you need to apply multiple rules.
```
ikoevska: just if or just when :)

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+    margin: 10;
+    padding: 10;
+}
+```
+
+<div class="exercise-end"></div>
+
+If you've done any web development before, the syntax should feel familiar here. You select four UI components (Page, TextField, Image, and Button) by their tag name, and then apply a handful of CSS rules as name/value pairs. NativeScript does not support all CSS properties because it is not possible to replicate some of them in native apps without causing performance issues. A [full list of the CSS properties that are supported](http://docs.nativescript.org/styling#supported-properties) are listed in the NativeScript docs.
+
+Let's make one more change. Although oftentimes you want CSS rules to apply equally to your iOS and Android app, occasionally it makes sense to apply a CSS rule to only one platform. For example, iOS text fields frequently have borders around them, but Android text fields do not. Let's look at how to make platform-specific style changes in NativeScript.
```
ikoevska: just often? :)

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+Next, add a `cssClass="link"` attribute to the sign up button in `login.xml`. The button's markup should look like this:
+
+``` XML
+<Button text="Sign up for Groceries" cssClass="link" />
+```
+
+<div class="exercise-end"></div>
+
+Let's break down what just happened. First, NativeScript supports CSS's `@import` statement for importing one CSS file into another. So this new line of code imports the CSS rules from `platform.css` into `app.css`. But, you might have noticed that Groceries does not have a file named `platform.css`—only `app/platform.android.css` and `app/platform.ios.css` exist. So what's going on here?
```
ikoevska: Here we have "So this new line... But... So..." Drop the last So and go with just "What's going on here?"

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+    background-color: transparent;
+}
+
+/* From platform.ios.css */
+.link {
+    border-width: 0;
+}
+```
+
+> **Tip**: NativeScript also supports selecting elements by the `id` attribute. Refer to the docs for [a full list of the supported selectors](http://docs.nativescript.org/styling#supported-selectors).
```
ikoevska: Capitalize tip

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+With these changes in place, you'll notice that the app looks halfway decent now, and also has a distinctly different look on iOS and Android:
+
+![login 1](img/cli-getting-started/chapter2/ios/3.png)
+![login 1](img/cli-getting-started/chapter2/android/3.png)
+
+Feel free to take some time to play with the look of this app before moving on. You can try adding some additional CSS class names, or adding some page-specific styles in your `login.css` file. When you're ready, let's move on and add an image to this login screen.
+
+### Images
+
+In NativeScript you use the `<Image>` UI component and its `src` attribute to add images to your pages. The `src` attribute allows you to specify your image in three ways. The first (and simplest) way is to point at the URL of an image:
```
ikoevska: lets you specify your image

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+In `login.xml`, add the `<Image>` below as the first child of the existing `<StackLayout>` tag:
+
+``` XML
+<Image src="res://logo" stretch="none" horizontalAlignment="center" />
+```
+
+<div class="exercise-end"></div>
+
+The `res://` syntax tells NativeScript to use a platform-specific resource, in this case an image. You might recall from before that platform-specific resources go in your app's `app/App_Resources` folder, and if you look there you'll find a few different image files, several of which are named `logo.png`.
```
ikoevska: Just drop the "You might recall from before part" and make it "Platform-specific resources go in your app's `app/App_Resources` folder. If you look there, you'll find a few..."

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+Although more complex than putting an image directly in the `app` folder, using platform-specific images gives you more control over image display on different device dimensions. For example iOS lets you use provide three different image files for devices with different pixel densities. As such you'll find logos named `logo.png`, `logo@2x.png`, and `logo@3x.png` in your `App_Resources/iOS` folder. For Android you'll find similar image files in `App_Resources/Android/drawable-hdpi` (for "high" dpi, or high dots-per-inch), `App_Resources/Android/drawable-mdpi` (for medium-dpi), and `App_Resources/Android/drawable-ldpi` (for low-dpi).
```
ikoevska: "iOS lets you use provide three different image files" -> either just use or just provide

ikoevska: I don't think you need to explain dpi here.

getting-started/chapter-2.md
```diff
@@ -0,0 +1,300 @@
...
+
+Once these files are in place the NativeScript framework knows how to pick the correct file; all you have to do is reference the image using `res://` and its base file name—i.e. `res://logo`. With this change in place here's what your login screen should look like on iOS and Android:
```
ikoevska: "in place" in consecutive sentences. Rephrase.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
+## Application logic
+
+In this chapter, you'll learn how to add JavaScript logic to your NativeScript app, and you'll be doing so using the base pattern on which the NativeScript framework is built, MVVM, or "model, view, view model". Here's what those words mean:
```
ikoevska: Hm, I don't think you need the commas in "model, view, view model" - https://en.wikipedia.org/wiki/Model_View_ViewModel

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+
+- **Model**: The model defines and represents the data. Separating the model from the various views that may use it allows for code reuse.
```
ikoevska: You need to stick with either may or might. I've been going with might for the AB docs.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+- **View**: The view represents the UI, which in NativeScript is written in XML. The view is often data-bound to the view model so that changes made to the view model in JavaScript instantly trigger visual changes to UI components.
+- **View Model**: The view model contains the application logic (often including the model), and exposes the data to the view. NativeScript provides a module called 'Observable', which facilitates creating a view model object that can be bound to the view. We'll discuss that shortly.
```
ikoevska: Remove this "We'll discuss that shortly."

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+> **NOTE**: The keyword `exports` is part of [CommonJS](http://wiki.commonjs.org/wiki/Modules/1.1), the standard on which both NativeScript and Node.js' implementations of modules are based. In CommonJS-based JavaScript modules, a free variable called `exports` is an object to which a module may add properties and methods to configure its external API. Using `exports` in a code-behind file exposes the function for use in the view, or XML file. That is, the `exports.load` assignment in the code-behind file is what makes `loaded="load"` in the view work.
+
+When you run the app with this change, NativeScript triggers the `load()` function you created in the code-behind file, and you should see the word “hello” logged in your terminal.
+
+![](img/cli-getting-started/chapter3/terminal-1.png)
+
+This simple example shows you how you can append attributes to UI components to run functions in the view's accompanying JavaScript file. Let's use another one of these attributes: `tap`.
+
+<h4 class="exercise-start">
+    <b>Exercise</b>: Add tap attributes to the login buttons and add their functions
```
ikoevska: does their functions stand for the tap attributes or for the login buttons? I suggest rephrasing this to drop the second "add" and to make it more clear.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+At this point, if you run your app and tap either of the buttons, you will see the appropriate alerts pop up. 
+
+![login 5](img/cli-getting-started/chapter3/ios/1.png)
+![login 5](img/cli-getting-started/chapter3/android/1.png)
+
+Now that you can see tap gestures working, let's make them do something more interesting than open alerts.
+
+### Navigating screens
+
+When you tap the “Sign Up for Groceries” button, you would expect a navigational change to a registration screen. This is very easy to do in NativeScript.
```
ikoevska: When you tap "Sign Up for Groceries"...

Also I am not sure whether the button name was Sign Up or Sign up. Please, check.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+exports.register = function() {
+    var topmost = frameModule.topmost();
+    topmost.navigate("./views/register/register");
+};
+```
+<div class="exercise-end"></div>
+
+This function uses the [frame module](http://docs.nativescript.org/ApiReference/ui/frame/README.html), which is the NativeScript module responsible for navigation in your app. Here, you tell the topmost frame, or the frame the user actually sees, to navigate to the register view. 
+
+If you run your app and click the “Sign up for Groceries” button, you will be sent to the registration screen, which we have pre-built for you.
```
ikoevska, line 103: Yep, another instance with different spelling for this button. Please, keep it consistent.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+![navigate](img/cli-getting-started/chapter3/ios/2.gif)
+![navigate](img/cli-getting-started/chapter3/android/2.gif)
+
+Now that you can access the registration page, go ahead and sign up for an account to use for the rest of this tutorial.
+
+<h4 class="exercise-start">
+    <b>Exercise</b>: Register for an account
+</h4>
+
+Open Groceries and click on the “Sign up for Groceries” button to access the registration page. From there, provide an email address and password, and click the “Sign Up” button to create an account.
```
ikoevska: ...tap "Sign up for Groceries to access...

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+
+Feel free to use a fake email address and “password” for your password, just remember to write down the credentials you enter as you'll need them later.
```
ikoevska: You can use a fake email address and password, just remember your credentials because you'll need them later.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+
+<div class="exercise-end"></div>
+
+> **TIP**: Although our Groceries app doesn't use complex navigation strategies, you have several available to you out of the box, such as the [TabView](http://docs.nativescript.org/ui-views#tabview) and the [SegmentedBar](http://docs.nativescript.org/ui-views#segmentedbar). A SideDrawer component is also available for free via Telerik's [UI for NativeScript](https://www.nativescript.org/blog/welcome-to-telerik-ui-for-nativescript) product.
+
+### Accessing UI components
+
+Since you have been working with forms in NativeScript, you need to understand how data flows back and forth between the front end and back end. Specifically, you have created a login form, but you need to be able to know what the user types into the fields in order to log them in to your app.
```
ikoevska: This is ambiguous - you can understand it either as "from when you started working with forms" or as "because you are working with forms". I suggest rephrasing. You can go with "Having been working with forms in NativeScript for a while now, you need to understand how..." or the even simpler "It's time to see how data flows back and forth between the front end and back end in forms."

Frankly, the entire paragraph needs a bit of rework.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+```
+
+With an `id` attribute in place, you can access this text field in your code-behind file. To do that, start by opening `app/views/login/login.js` and adding the code below at the top, underneath the `frameModule` variable.
+
+``` JavaScript
+var viewModule = require("ui/core/view");
+var email;
+```
+
+> **NOTE**: This line of code imports the [view module](https://docs.nativescript.org/ApiReference/ui/core/view/View.html), which is the base class for all UI components. The module provides a lot of functionality, including the ability to control properties of a view and its children. You're going to use it to get access to the email text field.
```
ikoevska: Why don't you make these links testable on testdocs.nativescript.org by making them relative?

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+    var page = args.object;
+    page.bindingContext = user;
+};
+```
+
+<div class="exercise-end"></div>
+
+What's going on here?
+
+- First, you're creating a `user` view model that is based on the NativeScript Observable module. You create the view model with two properties, `email` and `password`, that are pre-populated with some dummy values.
```
ikoevska: You can probably go with a numbered list here, dropping the first and next in the beginning of the two list items.

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+- Next, you bind the page to the `user` view model by setting it as the page's `bindingContext` property. This is specifically what makes the curly bracket syntax work.
+
+Simply put, properties placed on a page's binding context are available to XML elements using the `{% raw %}{{ propertyName }}{% endraw %}` syntax. Because JavaScript sets the view model's `email` to `"user@domain.com"`, and because you bound the email address text field to that property using `<TextField text="{% raw %}{{ email }}{% endraw %}">`, when you run this app you'll see "user@domain.com" appear on the front end.
+
+![](img/cli-getting-started/chapter3/ios/3.png)
+![](img/cli-getting-started/chapter3/android/3.png)
+
+What's really cool is that the binding is two-way. Meaning, when the user types text in these text fields, those changes are immediately made to your view model.
```
ikoevska: are immediately applied

getting-started/chapter-3.md
```diff
@@ -0,0 +1,224 @@
...
+
+In order to use these values, and to make this login functional by tying the your app into a backend service, you're going to need the ability to make HTTP calls. And to make HTTP calls in NativeScript you use the NativeScript http module. Let's look at how NativeScript modules work.
```
ikoevska: To use these values...

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
+## NativeScript modules
+
+In this chapter you'll learn about NativeScript modules, which are the JavaScript modules in your app's `tns_modules` folder. Whether you've realized it or not, you've already used several NativeScript modules. This includes the modules you've brought in via `require()`—the view module, the frame module, the observable module, and so forth—and also the UI components you've been using in XML, as `<Page>`, `<Image>`, `<TextField>`, and `<Button>` are all just NativeScript modules.
```
ikoevska: The "this includes the modules" sentence is way too long and somewhat unclear (especially if you're scanning through it.) Consider simplifying it. E.g. "This includes the modules you've brought in via `require()` (the view, frame and observable modules) and also the UI components you've been using in XML (the page, image, text field and button modules).

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+
+If you dig into `tns_modules` you can get an idea of how these modules work. Start by finding the `app/tns_modules/http` folder, which includes the implementation of the http module. It includes:
+
+- a `package.json` file that sets the name of the module;
+- a file containing the module's Android implementation (`http-request.android.js`);
+- a file containing the module's iOS implementation (`http-request.ios.js`);
+- a generic file (`http.js`) that abstracts the platform-specific code above into a platform-agnostic API, so that you can make HTTP calls on both platforms using a single JavaScript API.
```
ikoevska: drop "above"

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+
+> **NOTE**: You can refer to the [Node.js documentation on folders as modules](https://nodejs.org/api/modules.html#modules_folders_as_modules) for more detailed information on how NativeScript organizes its modules. (Remember that NativeScript modules adhere to the same CommonJS specification that Node.js modules do.)
```
ikoevska: remove the remember part

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+
+Finally, replace the `exports.signIn()` function with the code below:
+
+``` JavaScript
+exports.signIn = function() {
+    user.login();
+};
+```
+
+> **TIP**: You can always view the completed codebase in the 'master' branch of the [sample-Groceries repo](https://github.com/NativeScript/sample-Groceries.git).
```
ikoevska, line 126: Yay! :tada:

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+};
+```
+<div class="exercise-end"></div>
+
+This code handles both a successful and unsuccessful login. On success, you call the frame module's `navigate()` method to navigate the user to the (currently empty) list page. On failure, you use the dialog module to show the user an error message. Try inputting some invalid credentials to see what the dialog looks like.
+
+![login 8](img/cli-getting-started/chapter4/ios/1.png)
+![login 8](img/cli-getting-started/chapter4/android/1.png)
+
+With that, the login page is completely functional. Now that you have user management working in your NativeScript app, let's move onto the page that users will use to manage their grocery list. To do that, you need a module that shows items in a list, which is exactly what the ListView module does.
```
ikoevska: "the page where users will manage their grocery list."

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+> ``` JavaScript
+> var user = new UserViewModel({
+>     email: "username@domain.com",
+>     password: "password"
+> });
+> ```
+
+### ListView
+
+As its name implies, the ListView widget lets you show a list of things on the screen, which is exactly what you need for showing a list of groceries. Before tying the grocery list to a backend API, let's start by seeing how to show a hardcoded list of items on the screen.
```
ikoevska: Delete "As its name implies"

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+            </ListView.itemTemplate>
+        </ListView>
+    </GridLayout>
+</Page>
+```
+<div class="exercise-end"></div>
+
+> **NOTE**: Notice that this page is going to use a `<GridLayout>` to layout the UI components on the screen. As you add more UI components, you'll start dividing the screen into rows and columns, but for now you're just going to let the `<ListView>` take up the full screen (which is the default behavior of a `<GridLayout>` with no attributes).
+
+As discussed earlier, even though you're using `<ListView>` in XML, the ListView module is still a NativeScript module. You can find its implementation in the `tns_modules/ui/list-view` folder. If you want to, you could construct a ListView in pure JavaScript code in the code-behind file as shown in [this example](http://docs.nativescript.org/ApiReference/ui/list-view/HOW-TO.html). For most situations using the NativeScript UI modules in XML is easier, so we'll be sticking with XML usage throughout this tutorial.
```
ikoevska: Remove "As discussed earlier"

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+
+One thing to note here is the use of `<ListView.itemTemplate>`. This tag gives you the ability to control how each of the ListView's items displays within the list. For now you're using a simple `<Label>` UI component to display the `{% raw %}{{ name }}{% endraw %}` of each grocery.
```
ikoevska: Note the use of `<ListView.itemTemplate>`. This tag lets you control...

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+viewModel.empty = function() {
+    while (viewModel.length) {
+        viewModel.pop();
+    }
+};
+```
+
+<div class="exercise-end"></div>
+
+The code to make an HTTP call should look familiar, as it's using the same http module you used in the previous section. Here, you're using the http module's `getJSON()` method, which automatically takes care of formatting the backend's response as JSON.
```
ikoevska: Way too many use-based words in two sentences :)

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+
+But the most important thing to note here is the use of the `row` and `col` attributes. These attributes are zero-based, so the text field's `row="0" col="0"` attributes place it in the first row and first column, whereas the button's `row="0" col="1"` attributes place it in the first row and second column.
+
+Finally, replace the `<ListView>` tag with the code below to place it in the second row, to have it span both columns, and also to give it an `id` attribute that you'll need later.
+
+``` XML
+<ListView items="{% raw %}{{ groceryList }}{% endraw %}" id="groceryList" row="1" colSpan="2">
+```
+
+Now you just need to make the necessary changes to the code behind file to support these XML changes. Open `list.js` and start by adding a new `"grocery"` property to the `pageData` Observable. The `pageData` assignment should look like this:
```
ikoevska: code-behind

getting-started/chapter-4.md
```diff
@@ -0,0 +1,473 @@
...
+    <b>Exercise</b>: Implement the ability to add groceries to the list
+</h4>
+
+Open `app/views/list/list.xml` and change the `<GridLayout>` tag to use the code below:
+
+``` XML
+<GridLayout rows="auto, \*" columns="2\*, \*">
+```
+
+The `rows` attribute divides the screen into two rows, the first auto-sized according to its childrens' height, and the other to contain *, or the remaining height of the screen. The `columns` attribute divides the screen into two columns, where the first column takes up 2/3 of the screen, and the second takes up the remaining third.
```
ikoevska: Escape the asterisk - it messes up your formatting, not to mention that it disappears from the processed HTML. You can escape it like this `\*`

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
+## Plugins and npm modules
+
+Often, you need functionality that is not implemented in the NativeScript modules. But no worries, as NativeScript allows you to leverage [npm](https://www.npmjs.com/) (node package manager) to import npm modules into your apps. Alternately, you can install NativeScript plugins, which are simply npm modules that can access native code and leverage Android and iOS SDKs, if required.
```
ikoevska: lets you leverage

ikoevska: There are two leverage-s in two consecutive sentences. Rework this.

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
+## Plugins and npm modules
+
+Often, you need functionality that is not implemented in the NativeScript modules. But no worries, as NativeScript allows you to leverage [npm](https://www.npmjs.com/) (node package manager) to import npm modules into your apps. Alternately, you can install NativeScript plugins, which are simply npm modules that can access native code and leverage Android and iOS SDKs, if required. 
+
+In this chapter, you'll install and use an external module, an email validator, so that you can check email addresses for validity as they are entered on the registration screen. Then, you'll add a NativeScript plugin, the [NativeScript social share widget](https://www.npmjs.com/package/nativescript-social-share), to allow users to share their grocery lists using their device's native sharing widget.
```
ikoevska: You will use an email validator to check email address for validity sounds redundant. Try reworking this part. probably "In this chapter, you'll install and use and external email validator module to verify the format of email addresses as they are entered on the registration screen."

ikoevska: to let users share their grocery lists

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
...
+    <b>Exercise</b>: Use the email validator module
+</h4>
+
+Open `/app/shared/view-models/user-view-model.js` and add the following line at the top of the file:
+
+``` JavaScript
+var validator = require("email-validator/index");
+```
+
+> **NOTE**: You need to explicitly point to the validator module's `index.js` file. The NativeScript framework is configured to look at the `"main"` value in an npm module's `package.json` file. In the case of this particular module, the `"main"` value is simply `"index"` (rather than `"index.js"`) so you need to reference the index file directly. If this package's `"main"` were set to `"index.js"`, you could use the simpler `require("email-validator")`.
```
ikoevska: in the case of this module

getting-started/chapter-0.md
tjvantoll: done

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
...
+```
+
+<div class="exercise-end"></div>
+
+In this function, the user submits an email and password, and the value is sent to the view model for validation. If it passes, registration can proceed, otherwise you show an alert:
+
+![](img/cli-getting-started/chapter5/ios/1.png)
+![](img/cli-getting-started/chapter5/android/1.png)
+
+In general npm modules greatly expand the number of things you're able to do in your NativeScript apps. Need to do date and time formatting? Use [moment](https://www.npmjs.com/package/moment). Need utility functions for objects and arrays? Use [lodash](https://www.npmjs.com/package/lodash) or [underscore](https://www.npmjs.com/package/underscore). This code reuse benefit gets even more powerful when you bring NativeScript plugins into the picture.
```
ikoevska: Need date and time formatting?

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
...
+<Page.actionBar>
+    <ActionBar title="Groceries">
+        <ActionBar.actionItems>
+            <ActionItem text="Share" tap="share" ios.position="right" />
+        </ActionBar.actionItems>
+    </ActionBar>
+</Page.actionBar>
+```
+
+This code defines an [ActionBar](https://docs.nativescript.org/ApiReference/ui/action-bar/ActionBar), which is a UI component that includes various menu items, enclosed in the `<ActionBar.actionItems>` tag. The `title` of the ActionBar allows you to show page-specific titles.
```
ikoevska: lets you show page-specific titles.

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
...
+    var listString = list.join(", ").trim();
+    socialShare.shareText(listString);
+};
+```
+
+<div class="exercise-end"></div>
+
+This code takes the groceries from the grocery list view model, converts the data into a comma-separated string, and passes that string to the social share module's `shareText()` method.
+
+Now when you run the app, you'll see a Share button at the top of the screen, which lets you share the list of groceries via the native iOS and Android sharing widgets.
```
ikoevska: You will see a Share button, which lets you share ... via the ... sharing widgets. Redundant much? Although, frankly, I can't think of a way to rephrase this right now...

getting-started/chapter-5.md
```diff
@@ -0,0 +1,175 @@
...
+
+![Social sharing widget on iOS](img/cli-getting-started/chapter5/ios/2.gif)
+![Social sharing widget on Android](img/cli-getting-started/chapter5/android/2.gif)
+
+Pretty cool, huh? The ability to use npm modules greatly expands the number of things you're able to do in a NativeScript app. Need to compose emails in your app? Try out the [NativeScript email plugin](https://www.npmjs.com/package/nativescript-email). Need to use the clipboard in your app? Try out the [NativeScript clipboard plugin](https://www.npmjs.com/package/nativescript-clipboard).
+
+If you're looking for NativeScript plugins start by doing a [search for “NativeScript” on npm](https://www.npmjs.com/search?q=nativescript). If you don't find the plugin you need, you can [request the plugin on our ideas portal](https://nativescript.ideas.aha.io/), or you can take a stab at [creating the plugin yourself](https://docs.nativescript.org/plugins).
```
ikoevska: start by searching for NativeScript on npm.

getting-started/chapter-6.md
```diff
@@ -0,0 +1,190 @@
+## Accessing native APIs
+
+The beauty of NativeScript is that you can write a native iOS or Android app in JavaScript, XML, and CSS without touching Swift, Objective-C, or Java, if you choose. But what if you want to present a different, more platform-specific UI to your users? Or if you want to access an iOS or Android API that NativeScript doesn't expose through a NativeScript module or plugin?
+
+NativeScript gives you the option to dig into native code as needed, and to do so without leaving JavaScript. To show how this works in action, let's start by adding some color to the ActionBar in the iOS version of the Groceries app.
+
+### Customize the ActionBar - iOS
+
+When you use the ActionBar UI component in a NativeScript app, NativeScript is actually creating and managing an iOS `UINavigationController` for you. You can confirm this by digging into the implementation code, specifically `tns_modules/ui/action-bar/action-bar.ios.js`, and `tns_modules/ui/frame/frame.ios.js`.
```
ikoevska: When you use the ActionBar UI component, NativeScript is...

getting-started/chapter-6.md
```diff
@@ -0,0 +1,190 @@
...
+    navigationBar.barTintColor = UIColor.colorWithRedGreenBlueAlpha(0.011, 0.278, 0.576, 1);
+    navigationBar.titleTextAttributes = new NSDictionary([UIColor.whiteColor()], [NSForegroundColorAttributeName]);
+    navigationBar.barStyle = 1;
+    navigationBar.tintColor = UIColor.whiteColor();
+}
+```
+
+<div class="exercise-end"></div>
+
+Ok, let's break down what just happened, starting with the `if (page.ios)` check. NativeScript modules, in general, follow the pattern of exposing their native implementations through `ios` and `android` properties. You can see this in the if check (`page.ios`), and also on the first line within the if check, as `frameModule.topMost().ios` is used to retrieve a reference to the underlying `UINavigationController`. Testing for the existence of these properties (e.g. `if (page.ios)`) is a convenient way to fork your code, to ensure that iOS-specific code only runs on iOS, and Android-specific code only runs on Android.
```
ikoevska, line 49: Right now, as a user, I'm wondering - hey, why am I not doing this in login.ios.js and login.android.js. Just saying :)

getting-started/chapter-6.md
```diff
@@ -0,0 +1,190 @@
...
+        <Image src="res://ic_menu_delete" ios:visibility="collapsed" col="1" tap="delete" />
+    </GridLayout>
+</ListView.itemTemplate>
+```
+
+<div class="exercise-end"></div>
+
+With this code you're primarily adding an `<Image>` to the existing ListView template. But now that you have multiple UI components in this template, you have to tell NativeScript how to layout these two components, which is what the `<GridLayout>` handles. By specifying `columns="*, auto"` you divide each item into two columns: the first containing the label and the second containing the new image.
+
+For the image itself, the `ios:visibility="collapsed"` attribute sets the image's `visibility` CSS property to `"collapsed"`, which hides it. Because the attribute was prefixed with `ios:`, that CSS property is only applied on iOS; therefore the button displays on Android devices, but not on iOS ones. The trash can image itself has already been placed in the app for you, and can be found in appropriate sizes in the four drawable folders in `/app/App_Resources/Android/drawable-*dpi`. Here's what the trash can UI looks like on Android:
```
ikoevska: ...in the four drawable folders in `/app/App_Resources/Android/drawable-*dpi` is incorrect. You probably need to make this "in the four drawable folders in `/app/App_Resources/Android`"

getting-started/chapter-6.md
```diff
@@ -0,0 +1,190 @@
...
+            console.log(error);
+            reject();
+        });
+    });
+};
+```
+
+<div class="exercise-end"></div>
+
+This code probably looks fairly familiar by now. You're again calling the http module's `request()` method, this time specifying a `method` of `"DELETE"` to delete a grocery from the backend. You again return a `Promise` so the calling function has the opportunity to handle successful and unsuccessful calls. Note again the power of using the MVVM approach of building your app. To update the grocery list UI, all you have to do is remove the item from the ObservableArray (`viewModel.splice(index, 1)`), and let the list's presentation take care of itself.
```
ikoevska: "approach for building" probably?

getting-started/chapter-6.md
```diff
@@ -0,0 +1,190 @@
...
+
+If you run your app on Android you should have the ability to delete items from the list.
```
ikoevska: you should be able to delete items

getting-started/chapter-6.md
```diff
@@ -0,0 +1,190 @@
...
+
+The code to do so is a little trickier than our previous example for Android, so we've provided a ready-built module to implement this functionality in the `shared/utils/ios-swipe-delete.js` file. In this file you'll find a custom implementation of a data source that adheres to the `UITableViewDataSource` protocol. The file exports a single `enable()` function, which takes a reference to a ListView, and injects that ListView with the custom data source behavior.
+
+Don't worry too much about exactly what this code is doing, as it involves a bit of understand of how iOS APIs works. What is cool is that you *can* implement this relatively advanced iOS API in a few dozen lines of code, and that it's really easy to wrap this code with a very simple JavaScript API. This ease of use is exactly why NativeScript modules and NativeScript plugins are so easy to consume. Let's look at how to use this module.
```
ikoevska: "it involves a bit of understanding" probably?

getting-started/chapter-0.md
jlooper: updated, prefer the second option

getting-started/chapter-0.md
jlooper: updated, and noted

getting-started/chapter-1.md
jlooper: I'm going to leave the reference to AB in here for now, but I bet we will have to revisit this if a rebranding is in the works.

getting-started/chapter-1.md
jlooper, line 15: I think it's ok, leaving as  is.

getting-started/chapter-1.md
jlooper, line 57: added to chapter 7

getting-started/chapter-1.md
jlooper, line 59: I think I don't want to disrupt the flow, as I want to explain what this 'start' branch is before asking people to switch to it.

getting-started/chapter-1.md
jlooper: changed to IMPORTANT

getting-started/chapter-1.md
jlooper, line 93: I don't think so, this is a phrase describing the completion of the platform initialization

getting-started/chapter-1.md
jlooper, line 115: noted

getting-started/chapter-1.md
jlooper: removing that bit

getting-started/chapter-1.md
jlooper: true, I changed to 'a good way'

getting-started/chapter-1.md
jlooper: updated

getting-started/chapter-1.md
jlooper: updated

getting-started/chapter-2.md
jlooper, line 19: I added a note to this effect

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper, line 85: going to leave as is.

getting-started/chapter-2.md
jlooper: noted, and agree. Editing master sample code as well.

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: got it, thanks

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: 'oft' :) updated

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-2.md
jlooper: I'm going to disagree, because the way images are handled cross-platform can get really complicated and I want to elucidate a little in this area.

jlooper: updated the previous issue

getting-started/chapter-2.md
jlooper: updated

getting-started/chapter-3.md
jlooper: hm ok! updated

getting-started/chapter-3.md
jlooper: changed in 2 places, updated

getting-started/chapter-3.md
jlooper: updated

getting-started/chapter-1.md
ikoevska, line 93: Hm, you're right. I guess I've read this as "when the platform initialization completes". Keep it as it is.

getting-started/chapter-3.md
jlooper: changed to just 'Enable the buttons'

getting-started/chapter-2.md
ikoevska: About the dpi, I meant that you don't need to explain the abbreviation in the (for high dpi, or high dots-per-inch). You can just use dpi :)

getting-started/chapter-3.md
jlooper: it's Sign up. Edited

getting-started/chapter-3.md
jlooper, line 103: got it, updated

getting-started/chapter-3.md
jlooper: edited

getting-started/chapter-3.md
jlooper: edited

getting-started/chapter-3.md
jlooper: I chopped out half the cruft there. No need for it.

getting-started/chapter-3.md
jlooper: changed to relative:  `[view module](/ApiReference/ui/core/view/View.html)` - hope this is correct

getting-started/chapter-3.md
jlooper: edited

getting-started/chapter-3.md
jlooper: edited

getting-started/chapter-3.md
jlooper: edited

getting-started/chapter-4.md
jlooper: edited

getting-started/chapter-4.md
jlooper: edited

getting-started/chapter-4.md
jlooper: edited

getting-started/chapter-4.md
jlooper, line 126: :)

getting-started/chapter-4.md
jlooper: edited

getting-started/chapter-4.md
jlooper: edited

getting-started/chapter-4.md
jlooper: edited

getting-started/chapter-4.md
jlooper: agreed and edited. Using my keyboard lol.

getting-started/chapter-4.md
jlooper: edited in 2 places

getting-started/chapter-4.md
jlooper: got it this time, edited

getting-started/chapter-5.md
jlooper: edited both

getting-started/chapter-5.md
jlooper: edited both

getting-started/chapter-5.md
jlooper: edited

getting-started/chapter-5.md
jlooper: edited

getting-started/chapter-5.md
jlooper: edited

getting-started/chapter-5.md
jlooper: me neither, we'll have to suffer thru :dancer:

getting-started/chapter-5.md
jlooper: edited

getting-started/chapter-6.md
jlooper: edited

getting-started/chapter-6.md
jlooper, line 49: hmmmm. I see what you mean, but a little reluctant to offer alternatives and confuse people...

getting-started/chapter-6.md
jlooper: right, edited

getting-started/chapter-6.md
jlooper: edited

getting-started/chapter-6.md
jlooper: Boy oh boy, do I write in a verbose fashion. I do not understand why I write in such a verbose fashion. In order to close this note, I edited it.

getting-started/chapter-6.md
jlooper: yep, edited

tjvantoll:
Ok I believe all comments have been addressed now. I also pushed a few small edits per a review from Niki.

getting-started/chapter-6.md
ikoevska, line 49: My concern is more on the side that we offer too many alternatives for providing platform-specific settings and content and this might be confusing as they are scattered around the guide. In the beginning, we mention the platform-specific files and how the CLI processes them. Next, we have the if-statements in a JS file. And then, below, we have the platform-specific properties in XML. While all of these approaches are great and can be used in all scenarios, it might be a good idea to point to an article explaining the various scenarios in which it is best to use one approach or the other. Or to say - here are your options, however, it is a best practice to...

getting-started/chapter-5.md
ikoevska: Ok, since it's pretty obvious that the Share button lets you share things, probably something like "Now when you run the app, you'll be able to share groceries via the new button at the top of the screen. When you tap it, the native iOS and Android sharing widgets will pop up."

OR

"Now when you run the app, you'll see a new button at the top of the screen. When you tap it, the native iOS or Android sharing widget will show to let you post your groceries to your social networks or send them via email, message or any other method you prefer."

Or something like that.

getting-started/chapter-1.md
ikoevska: Yeah, my point exactly - we will need to revisit this once we drop the product names and go with the new design/code/backend design which promotes the Platform not as a suite of products and services but as a standalone product. This is why I am inclined to start referring to the Platform instead of AppBuilder in docs from now on. (Moreover, the Platform provides {N} for most services now - BES, Analytics, AB, SB.)

getting-started/chapter-1.md
ikoevska, line 59: Hm, ok. But for me, the more natural flow is the other way around.  You can still move the `master` vs `start` description close to the git clone part and then do the directory switch.

getting-started/chapter-4.md
ikoevska: This comment has not been marked as updated =) Please, check if you want to address it. :)

getting-started/chapter-6.md
ikoevska: TJ and you both have wonderful and very unique writing style (verbose or not :D). =) Unfortunately, the deal with documentation is that you have to be more on the minimal utilitarian side - say as much as you can with as minimum simple words as you can :D. (Frankly, when I first started writing docs, I used to think that I have to use the most complex words, forms and phrasing I can come up with but it proved quite the opposite :D - thank God!)

ikoevska:
Frankly, I don't think I can provide any more meaningful comments if I do another review this or next week. By the third/fourth read, I usually become "blind" to some parts of the text and tend to miss obvious typos/errors. I think we should deploy and test this on testdocs.nativescript.org and give it a few iterations of testing.

getting-started/chapter-0.md
ikoevska: Could this change be lost? It doesn't appear here?

ikoevska:
One more thing that we might want to address - imho all links to docs.nativescript.org should be relative instead.

getting-started/chapter-1.md
tjvantoll: My immediate concern is that the {N} docs currently refer to AppBuilder all over the place, and this tip links to a guide with “AppBuilder” in its name. So if we change this wording I'd want to change the page this links to, and so on.

I agree that we should go through the docs and change “AppBuilder” --> “Platform”, but personally I'd prefer that to be done outside of the scope of the getting started guide.

getting-started/chapter-1.md
tjvantoll, line 59: I like them being closer too @ikoevska, but I think the current order is a necessity. Unless there's some git magic I'm unaware of, I don't think you can change branches without first navigating into the folder that contains the git repository.

getting-started/chapter-5.md
tjvantoll: Switched to your second sample. Although I added in an Oxford comma, because I'm a rebel, and because I think it helps with readability.

getting-started/chapter-0.md
tjvantoll: This was my bad screwing up a merge. The change is there now.

getting-started/chapter-1.md
ikoevska: Oh, boy. You're absolutely right! I was thinking over this flow from my SourceTree perspective and from the AppBuilder quirk that the repo contains a duplicate folder and never thought of that. 

Ignore my comment =)

getting-started/chapter-1.md
ikoevska, line 59: Oh, boy. You're absolutely right! I was thinking over this flow from my SourceTree perspective and from the AppBuilder quirk that the repo contains a duplicate folder and never thought of that.

Ignore my comment =)

getting-started/chapter-6.md
tjvantoll, line 49: I played for this for a while and had a hard time coming up with succinct wording. The ideal to me is just linking to another article at this point that can spell out the options in detail (as you suggest)... but that article doesn't exist yet. I might need to write that.

For now what do you think of adding the tip below after the reference paragraph?

> **TIP**: As a best practice, testing for a platform with an if check is best done when you have a small number of platform-specific changes to make. If your code-behind file contains entirely different code for iOS and Android, your code will likely more readable if you use NativeScript's `.android.js` and `.ios.js` naming convention—e.g. `login.ios.js` and `login.android.js`.

tjvantoll:
> One more thing that we might want to address - imho all links to docs.nativescript.org should be relative instead.

All links are now relative.

tjvantoll:
> Frankly, I don't think I can provide any more meaningful comments if I do another review this or next week. By the third/fourth read, I usually become "blind" to some parts of the text and tend to miss obvious typos/errors. I think we should deploy and test this on testdocs.nativescript.org and give it a few iterations of testing.

I'm in favor of deploying this as soon as is reasonably possible. It'd be nice to point people testing this guide to testdocs rather than the now out of date https://github.com/NativeScript/quick-start.

getting-started/chapter-6.md
ikoevska, line 49: As a best practice, testing for a platform with an if check is the way to go when you have a small number of platform-specific changes to make. If, on the contrary, you have big, entirely different chunks of code for iOS and Android, you might want to go with platform-specific code-behind files - e.g. login.ios.js and login.android.js.

(and make sure to link to the chapter and section with this platform-specific naming convention, you might need to insert an `<a id>` element in there for better linking.)

ikoevska:
You have my :+1:  for merging and deploying this as soon as possible. While fixing any technical issues that might come up from testing, we can also fix any other style/phrasing issues as well.

getting-started/chapter-6.md
tjvantoll, line 49: Done https://github.com/tjvantoll/docs/commit/8cb330f149cef6f00f57d7d944f517c092333fcd

tjvantoll:
> You have my :+1: for merging and deploying this as soon as possible. While fixing any technical issues that might come up from testing, we can also fix any other style/phrasing issues as well.

Done :)

I'll work on getting my Jekyll changes merged in so we can view this live.

PR #140, opened 2015-09-10
Update CLI sys requirements to >=0.10.35 <1.0.0

teobugslayer:


setup/ns-cli-setup/ns-setup-linux.md
```diff
@@ -17,7 +17,7 @@ On Linux systems, you can use the NativeScript CLI to develop only Android apps.
 ## System Requirements
 
 * Ubuntu 14.04 LTS
-* Node.js 0.10.26 or a later stable official release except Node.js 0.10.34
+* Node.js 0.10.35 or a later stable 0.x release
```
ikoevska, line 20: Is this 0.10.35 really the case?

ikoevska:
This unannounced, unofficial jump to 0.10.35 bothers me. Also, you need to fix the Node.js installation instructions which currently point to Node.js and say that what you get from the Install button is what you need. You need to point users to the downloads section of the Node.js site.

teobugslayer:
We talked in person and agreed to merge this PR in time for 1.3 release. We should update the change log to specify that we drop support for node < 0.10.35

rosen-vladimirov:
:+1:

PR #155, opened 2015-09-25
Windows setup changes (Gradle and ANDROID_HOME)

tjvantoll:
Ping @rosen-vladimirov @ligaz @ikoevska 

I created this PR to 1) add the `ANDROID_HOME` requirement, which seems to have been missing from this page, and 2) to switch the Gradle installation instructions to use `choco`.

For Gradle the `choco` installation is waaaay easier, and the [CLI's README](https://github.com/NativeScript/nativescript-cli/blob/master/README.md) already recommends `choco install gradle` (thus these two pages are out of sync). The one caveat is `choco install gradle` installs Gradle 2.2, and the current docs say 2.3+ is required. I just went through the installation on a fresh machine, and my Android builds are working with Gradle 2.2, but there's certainly a chance that I'm missing something. Let me know what you think.

### DIFF
--- a/setup/ns-cli-setup/ns-setup-win.md
+++ b/setup/ns-cli-setup/ns-setup-win.md
@@ -48,7 +48,7 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
         SETX JAVA_HOME "Path to the jdk* install folder" /M
         ```
 
-        For example: `SETX JAVA_HOME C:\Program Files\Java\jdk1.8.0_11 /M`
+        For example: `SETX JAVA_HOME "C:\Program Files\Java\jdk1.8.0_11" /M`
 1. Install [Gradle 2.3](https://docs.gradle.org/current/userguide/installation.html) or a later stable release.
     1. In the command prompt, run the following command.
 
@@ -82,7 +82,7 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
         SETX JAVA_HOME "Path to the Android SDK install folder" /M
         ```
 
-        For example: `SETX ANDROID_HOME C:\Program Files (x86)\Android\android-sdk /M`
+        For example: `SETX ANDROID_HOME "C:\Program Files (x86)\Android\android-sdk" /M`
     1. Restart the command prompt.
 1. Install the required Android SDKs and the Android Support Repository.
 
--- a/setup/ns-cli-setup/ns-setup-win.md
+++ b/setup/ns-cli-setup/ns-setup-win.md
@@ -49,14 +49,19 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
         ```
 
         For example: `SETX JAVA_HOME C:\Program Files\Java\jdk1.8.0_11 /M`
-1. Download and install [Gradle 2.3](https://docs.gradle.org/current/userguide/installation.html) or a later stable official release.
-1. If not present, add the following file path to the `PATH` system environment variable.
+1. Install [Gradle 2.2](https://docs.gradle.org/current/userguide/installation.html) or a later stable release.
+    1. In the command prompt, run the following command.
 
-	```
-	Path to the bin directory in the Gradle installation folder
-	```
+        ```Shell
+        choco install gradle
+        ```
+    1. If not present, add the following file path to the `PATH` system environment variable.
 
-	For example: `PATH=...;...;C:\tools\gradle-2.3\bin`
+        ```
+        Path to the bin directory in the Gradle installation folder
+        ```
+
+        For example: `PATH=...;...;C:\ProgramData\chocolatey\lib\gradle\tools\gradle-2.2.1\bin\`
 1. Install the [Android SDK](http://developer.android.com/sdk/index.html).
     1. In the command prompt, run the following command.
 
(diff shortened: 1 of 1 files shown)


setup/ns-cli-setup/ns-setup-win.md
```diff
@@ -70,13 +75,20 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
         Path to platform-tools directory in the Android SDK installation folder
         ```
 
-        For example: `PATH=...;...;C:\Users\MyUser\AppData\Local\Android\android-sdk\tools;C:\Users\MyUser\AppData\Local\Android\android-sdk\platform-tools`
+        For example: `PATH=...;...;C:\Program Files (x86)\Android\android-sdk\tools;C:\Program Files (x86)\Android\android-sdk\platform-tools`
```
tjvantoll, line 78: This change switches the path to be the location `choco` puts the Android SDK at.

setup/ns-cli-setup/ns-setup-win.md
```diff
@@ -42,21 +42,26 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
...
 
-	```
-	Path to the bin directory in the Gradle installation folder
-	```
+        ```
+        Path to the bin directory in the Gradle installation folder
+        ```
 
-	For example: `PATH=...;...;C:\tools\gradle-2.3\bin`
+        For example: `PATH=...;...;C:\ProgramData\chocolatey\lib\gradle\tools\gradle-2.2.1\bin\`
```
tjvantoll: This is where `choco` currently places the Gradle files.

ikoevska: After all, we are asking people to use choco (see the steps above). It is only fair that the path examples stick with choco as well.

tjvantoll:
@rosen-vladimirov @ligaz

Any concerns with these changes?

setup/ns-cli-setup/ns-setup-win.md
```diff
@@ -21,7 +21,7 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
 * (Optional) Chocolatey to simplify the installation of dependencies
 * JDK 8 or a later stable official release
 * Android SDK 22 or a later stable official release
-* Gradle 2.3 or a later stable official release
+* Gradle 2.2 or a later stable official release
```
ligaz: We are currently giving warning for anything below `2.3` when [doctor](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/doctor-service.ts#L7)-ing. @teobugslayer can you please elaborate if we are going to work correctly with 2.2?

setup/ns-cli-setup/ns-setup-win.md
ligaz: This shouldn't be necessary because the Chocolate package adds this to the PATH.

tjvantoll: If I remember correctly Chocolate didn't take care of that for me for whatever reason. Let me check again.

setup/ns-cli-setup/ns-setup-win.md
```diff
@@ -42,21 +42,26 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
...
         ```
 
         For example: `SETX JAVA_HOME C:\Program Files\Java\jdk1.8.0_11 /M`
-1. Download and install [Gradle 2.3](https://docs.gradle.org/current/userguide/installation.html) or a later stable official release.
-1. If not present, add the following file path to the `PATH` system environment variable.
+1. Install [Gradle 2.2](https://docs.gradle.org/current/userguide/installation.html) or a later stable release.
+    1. In the command prompt, run the following command.
+
+        ```Shell
+        choco install gradle
```
rosen-vladimirov: Maybe we should note that currently the latest available version of gradle that can be installed with choco is 2.2.1, which is not working for us, we need at least 2.3 as far as I know.
Also we have some code issues (when the version is with three numbers), which make the `tns doctor` command to fail, but I'll try to fix them :)

setup/ns-cli-setup/ns-setup-win.md
teobugslayer: We explicitly require gradle 2.3. We **do not** work with anything bellow

ikoevska:
:+1: from me once you've figured out the Gradle versioning stuff :)

ErjanGavalji:
Shall we temporarily point to installing the [unapproved 2.3 choco-gradle](https://chocolatey.org/packages/gradle/2.3)?

tjvantoll:
Sorry for the late replies on this one everyone. Even though it's unapproved, I verified that `choco install gradle -version 2.3` works, and adds the appropriate directory to the Windows PATH... so I updated this PR to use it.

Any final concerns @ligaz @ErjanGavalji  @rosen-vladimirov? Otherwise I'll merge this in.

setup/ns-cli-setup/ns-setup-win.md
```diff
@@ -70,13 +75,20 @@ On Windows systems, you can use the NativeScript CLI to develop only Android app
...
 
-        For example: `PATH=...;...;C:\Users\MyUser\AppData\Local\Android\android-sdk\tools;C:\Users\MyUser\AppData\Local\Android\android-sdk\platform-tools`
+        For example: `PATH=...;...;C:\Program Files (x86)\Android\android-sdk\tools;C:\Program Files (x86)\Android\android-sdk\platform-tools`
+    1. If not present, create the following environment variable from a command-prompt, started as an administrator.
+
+        ```
+        SETX JAVA_HOME "Path to the Android SDK install folder" /M
+        ```
+
+        For example: `SETX ANDROID_HOME C:\Program Files (x86)\Android\android-sdk /M`
```
teobugslayer: I sense missing double quotes around paths with spaces - can we double-check these, please!

tjvantoll: Good catch. I was following the earlier example of `JAVA_HOME`, which was also wrong. I just pushed a commit that puts double spaces around both.

tjvantoll: And you're right. I verified that without the quotes this command fails.

ErjanGavalji:
:+1:

tjvantoll:
Thanks all. I merged this in.

One more question: should we continue to dual maintain the setup docs both here and in https://github.com/NativeScript/nativescript-cli/blob/master/README.md? I noticed the two are now out of sync, and I was going to send a PR, but having two copies of the same information seems a bit silly.

PR #167, opened 2015-10-21
Data Binding article - restructured and rewritten

N3ll:


bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
 
 #Data Binding
 
-##Overview
+Data binding is the process of connecting application user interface (UI) to a data object (code). There are two different ways for propagating а change. It is possible to update the source code through the UI componets and vise versa - to reflect in the UI the changes from modifying the code.
```
ikoevska: - Avoid "there is/there are" phrasing.
- What are these three ways you have mentioned? If you don't intend to name them, remove this sentence.
- Don't overcomplicate and overexplain - try to simplify your last sentence.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
 
-Data binding is the process of connecting application user interface (UI) to a data object (business model). With a correct data binding settings and if data object provides proper notifications then when data changes application UI will reflect changes accordingly (source to target). Depending on settings and requirements there is a possibility to update data from UI to data object (target to source).
+> **source** is used as any object in the code, and **target** as any UI control (like TextField).
```
ikoevska: What is this? Consider introducing this block with a simple sentence or phrase.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
 
-> **source** is used as any business logic object, and **target** as any UI control (like TextField).
+* [Data Flow Direction](#data-flow-direction)
+* [Basic Binding Concepts](#basic-binding-concepts)
+* [Create a Binding](#create-a-binding)
+    * [TwoWay Binding in Code](#twoway-binding-in-code)
```
ikoevska: Two-Way

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+    * [Binding in XML](#binding-in-xml)
+* [Binding Source](#binding-source)
+	* [Biding to a Property](#biding-to-a-property)
+	* [Binding to an Event in XML](#binding-to-an-event-in-xml)
+	* [Binding to a Plain Object](#binding-to-a-plain-object)
+	* [Binding to a Parent Binding Context](#binding-to-a-parent-binding-context)
+* [Using Expressions for Bindings](#using-expressions-for-bindings)
+	* [Supported Expressions](#supported-expressions)
```
ikoevska: Lists which consist of one item are not really a list. It is best to avoid one-itemed lists.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
-##Basic data binding concepts
+##Data Flow Direction
 
-Generally almost every UI control (since all controls are created with data binding in mind) could be bound to a data object. However there are few restrictions for data binding to work out of the box.
+Part of the data binding settings is the way data flows. NativeScript data binding supports the following data transmissions.
 
-* Target object should be a successor of **Bindable** class. This is the case with all NativeScript UI controls.
-* Target property should be a **dependency property** in order to use data binding from target to source (or two way data binding). A plain property could be used if there is no need of **twoWay** binding.
-* Data (business) object should raise **propertyChange** event for every change in the value of the property.
+* **OneWay** - the default setting, guaranteeing the update of the target property, when a change is made in the source property. Contrary, a change in the UI will not be propagated back to the code. If such a change is made, it will stop the binding connection.
```
ikoevska: In this case, is "OneWay" the exact spelling of the property value, or is it the general term? If the general term, go with One-Way.

ikoevska:
Consider expanding the introductory section of this article. You might want to list API methods and properties associated with data binding, if any. Links to related API/runtime documentation will be helpful as well.

bindings.md
ikoevska: Go for the so called parallelism in lists - use similar structure for the items in your lists. In this case, consider going with "this setting is the default..."

In addition, for this sentence, consider simplifying it. As a general rule, it is best to go with simple structures as to avoid the confusion of the reader and to spare them the untangling of pretty but overly complicated structures (not this case). Try something along the lines of "this is the default setting which ensures that the target property updates when a change in the source property occurs". See how this works and try experimenting it.

ikoevska: As a general rule, avoid passive voice and go with active voice instead. Here, instead of "when a change is made" consider "when a change occurs" and pick the one that appeals more to you.

ikoevska: This "Contrary," here bothers me a bit. A "However," might be better here. Again, as a general rule, most of the time you don't need these words in your technical writing, so consider removing it and see how it sounds.

Again, try reworking the passive voice here and in the next sentence.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
 
-##Direction of data flow
+* **TwoWay** - this setting ensures the reflexion of changes in both directions - from target to source and source to target. It is very useful in the cases when user input should be handled.
```
ikoevska: We're using US spelling and grammar in general. Go with reflection.

ikoevska: As a general rule, we tend to avoid "It is useful", "it is recommended" and similar phrases. Consider "You can use two-way data binding when you need to handle user input."

ikoevska: And my remark for TwoWay vs two-way still stands here. :)

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+* **OneWay** - the default setting, guaranteeing the update of the target property, when a change is made in the source property. Contrary, a change in the UI will not be propagated back to the code. If such a change is made, it will stop the binding connection.
 
-##Direction of data flow
+* **TwoWay** - this setting ensures the reflexion of changes in both directions - from target to source and source to target. It is very useful in the cases when user input should be handled. 
 
-Part of the data binding settings is the way data (values) flows. NativeScript data binding supports following data transmissions.
+##Basic Binding Concepts
 
-* **OneWay** - this is a setting that indicates that a change in the source object will update target property, but a change in target property will not be propagated back to source (data object). Any update to the target property (except binding) will stop the binding connection. - this is the **default** option.
+Generally almost every UI control (since all controls are created with data binding in mind) could be bound to a data object. However, there are a few requirements to be fulfilled in order for the data binding to work out of the box.
```
ikoevska: Typically, it is a good idea to use because instead of due and since to avoid ambiguity. This is especially important when creating content for localization and for non-native speakers who might not be aware of the alternative meanings of due and since.

ikoevska: Rework this paragraph in a you-oriented active voice.

Actually, go through the entire article to rework passive voice instances. I will not mark them anymore.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
 
-* **TwoWay** - this setting indicates that both changes in source object will be reflected to the target and changes from target will update the source object. Very useful in cases when user input should be handled (for example user writes a text - text is written within a TextField control and is updated to a underlying data property of the source object).
-In order to use this option binding options should set **twoWay as true**. Following examples show how to do that.
+* The target object should be a successor of the **Bindable** class. All NativeScript UI controls already inherit from this class.
```
ikoevska: Consider using a very limited set of modal verbs to avoid the confusion caused by the various nuances. For most of the cases "must", "have to" and "can" will be more than enough.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+* For **twoWay** data binding, the target property should to be a **dependency property**. For **oneWay** binding, using a plain property is sufficient.
```
ikoevska: Move the second sentence to a separate bullet.

Also if there are additional documentation resources for "dependency property", link to them.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+* The data object should raise a **propertyChange** event for every change in the value of its property in order to notify all of the listeners, interested in the change.
```
ikoevska, line 43: Is this data object a target object or a source object or it doesn't matter? Link to any documentation for raising a propertyChange event.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
 
-##Creating a binding
+##Create a Binding
 
-* Creating binding in code.
+###TwoWay Binding in Code
```
ikoevska: Two-Way for the general term, twoWay for the code setting. Apply as needed throughout (including for one-way).

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+The example bellow consists of a Label, TextField and a source property to which the UI controls are bound. Our purpose will be, when the user enters an input in the TextField, the property in the code to be updated together with the Label text. We will bind the TextField with a twoWay binding, so the user input will change the property in the code. And the binding of the Label will be set to oneWay in order to propagate changes only from the code to the UI. But, as we already mentined, when we make a change from the UI (TextField) to the source property, the binding will stop. This means that we will not have a constant flow of progating changes from the source property to the Label. Thus the property in the code should raise a **propertyChange** event, in order to notify the Label for the changes. To raise this event, we will use an inbuild class, which provides this functionality - Observable.
```
ikoevska: Below ([bellow](http://www.merriam-webster.com/dictionary/bellow) is something else entirely).

ikoevska: I have a huge dislike for "our/we" approach to documentation. I find it welcome in blogs but not in docs. However, going with an "our/we" approach is something that we might need to consider for the NativeScript documentation. If we decide to use it, we should go all the way and apply it to the entire documentation set. Note this down for further discussion.

ikoevska: Code formatting to code elements.

ikoevska: Consider dispersing this very lengthy explanation throughout the code block below as code comments. Your readers are more likely to pay attention to the code comments rather than this lengthy paragraph.

ikoevska: inbuild, inbuilt or built-in class?

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+source.textSource = "Text set via twoWay binding";
 ```
 ``` TypeScript
 import observableModule = require("data/observable");
 var source = new observableModule.Observable();
+source.textSource = "Text set via twoWay binding";
 ```
 
-Creating a target object. For the sake of example we will also create a target object an instance of Bindable class (all UI controls derives from it).
+Next we need to create **target** objects to bind to the source property. In our case these will be a Label and a TextField, which inherit from the Bindable class (as all of the UI controls do).
```
ikoevska: Comma after next

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
 
-* **TwoWay** - this setting indicates that both changes in source object will be reflected to the target and changes from target will update the source object. Very useful in cases when user input should be handled (for example user writes a text - text is written within a TextField control and is updated to a underlying data property of the source object).
-In order to use this option binding options should set **twoWay as true**. Following examples show how to do that.
+* The target object should be a successor of the **Bindable** class. All NativeScript UI controls already inherit from this class.
+* For **twoWay** data binding, the target property should to be a **dependency property**. For **oneWay** binding, using a plain property is sufficient.
+* The data object should raise a **propertyChange** event for every change in the value of its property in order to notify all of the listeners, interested in the change.
 
-##Creating a binding
+##Create a Binding
```
ikoevska: It is unclear that this is an example section.

bindings.md
```diff
@@ -7,92 +7,121 @@ position: 6
...
+	twoWay: false
+};
+targetLabel.bind(labelBindingOptions, source);
 ```
 
-* Create a data binding in xml
+###Binding in XML
 
-	1. Create a source object. "source" object from the previous case (creating binding with code) will be used for following examples.
+To create a binding in XML, we will need a source object, which will be created the same way, as in the exmple above ([TwoWay Binding in Code](#twoway-binding-in-code)). Then we describe the binding in the XML (using a mustache syntax). With an XML declaration we set only the names of the properties - for the target: text and for source: textSource.The interesting thing here is that we don't specify explicitly the source of the binding. More about this topic will be discussed in the next article.
```
ikoevska: Link to a definition of mustache syntax. Or better yet, consider creating a Glossary article in your docs with terms and phrases you are likely to use in the documentation but which could be considered NativeScript specific or a sort of a professional slang. Mustache syntax is a very good candidate for such a glossary.

ikoevska: Missing space between "source: textSource" and "The interesting thing here"

Also, what is this next article you're referring to? Name it and better yet - link to it.

Another general rule: never take for granted that users will be reading your article as a part of a set or that they will be aware of the sequence of your imagined set. Most likely, with the way users use the documentation - they would have landed on this article from Google and might have no idea that they must have read some previous article or that they must read some other article next. If there is such "recommended" or "required" reading, make sure to always point it out and link to it.

bindings.md
```diff
@@ -102,13 +131,13 @@ source.get("textSource");
 </Page>
 ```
 
-> Note: For an UI elements created with an xml declaration when data binding is set **twoWay** option is **true** by default.
-
-With an xml declaration we set only properties names both for target (text) and source (textSource). The interesting thing here is that there is no source of the binding (actually it is not set directly).
+> Note: For an UI elements created with an XML declaration when data binding is set **twoWay** option is **true** by default.
```
ikoevska: For UI elements or For a UI element. Consider simplifying your note - I had to read it two or three times to grasp it.

Btw, because I'm always wondering whether it's a UI or an UI - here's a [helpful resource](http://www.ecenglish.com/learnenglish/when-use-a) that explains it. (Let's hope I'll finally remember the rule :D)

bindings.md
```diff
@@ -102,13 +131,13 @@ source.get("textSource");
...
 
 ##Binding source
 
-The important part of the data binding is setting the source object. NativeScript data binding works with any object that emits a **propertyChange** event. On the process of creating binding **source** can be set as second parameter of the bind(bindingOptions, source) or could be omitted. In the case (when source argument is omitted) for source is used a special property named **bindingContext** of the Bindable class. The special about this property is that it is inheritable across the visual tree. This means that control can use the **bindingContext** (as source) of the first **parent** element with a explicitly set **bindingContext**. With the previous example **bindingContext** can be set either on Page instance or StackLayout instance and TextField will have a proper **source** for its "text" property binding.
+###Biding to a Property
```
ikoevska: Biding or Binding?

bindings.md
```diff
@@ -102,13 +131,13 @@ source.get("textSource");
...
+###Biding to a Property 
+
+An important part of the data binding is setting the source object. For a continuous flow of data chnages, the source property needs to emit a **propertyChange** event. NativeScript data binding works with any object that emits it. To add a binding **source**, it could be set as a second parameter of the method **bind(bindingOptions, source)**. However, this parameter is optional and could be omited, in which case for source is used a special property named **bindingContext** of the Bindable class. What is special about this property is that it is inheritable across the visual tree. This means that a UI control can use the **bindingContext** (as source) of the first **parent** element with an explicitly set **bindingContext**. With the previous example **bindingContext** can be set either on Page instance or StackLayout instance and the TextField will have a proper **source** for the binding of its "text" property.
```
ikoevska: emits it -> it's a good idea to be more specific about the it here.

ikoevska: To add a binding **source**, it could be set as a second parameter of the method **bind(bindingOptions, source)**. -> Read this out loud. Doesn't it sound odd? Consider rephrasing.

ikoevska: Please, rework this paragraph. It might be due to the fact that I am quite the noob in NativeScript and programming in general but I got lost in all the "this property", "that property", "it" and so on. Also, too many "special"s.

bindings.md
```diff
@@ -120,11 +149,17 @@ page.bindingContext = source;
 //or
 stackLayout.bindingContext = source;
 ```
+###Binding to an Event in XML
 
-* Create a data binding to an event in xml
-
-There is an option to bind a function to execute on a specific event (MVVM command like). This option is available only through an xml declaration. The different part is that the source property should have an event handler function as value.
+There is an option to bind a function to execute on a specific event (MVVM command like). This option is available only through a XML declaration. To implement such a functionality, the source object should have an event handler function.
```
ikoevska: an XML declaration

bindings.md
```diff
@@ -137,22 +172,92 @@ source.set("onTap", function(eventData) {
 });
 page.bindingContext = source;
 ```
+> Note: Be aware that if there is an event handler function **onTap** within the page code behind ([more info about xml declarations](./ui-with-xml.md)), and **onTap** function within the **bindingContext** object then there **will not** be 2 event handlers hooked for that button. If the function in the code behind should be execute, the following syntax should be used in the XML - **tap="onTap"** and if the function form the bindingContext - **tap="\{\{ onTap \}\}"**.
```
ikoevska: - In this case -> code-behind.
- Capitalize XML in "xml declarations".
- Generally, numbers from 0 through 9 are written as words -> use two here.
- What is "if the function form the bindingContext"? I believe there's either a typo or something missing here.
- You should mention the button somewhen earlier.

bindings.md
```diff
@@ -137,22 +172,92 @@ source.set("onTap", function(eventData) {
...
+
+###Binding to a Plain Object
 
-and how xml will look like:
+Very common case is to provide a list (array) of plain elements (numbers, dates, strings) to a `listview` items collection. All examples above demonstrated how to bind an UI element to a property of the bindingContext. If there is only plain data there is no property to bind, so the binding should be to the entire object. Here comes in place another feature of NativeScript binding - object or value binding. To refer to the entire object, which is Date() in our example, the keyword `$value` should be used.
```
ikoevska: A very common case

ikoevska: All examples above demonstrate

They haven't stopped demonstrating what they're demonstrating.

For the most part, strive to use present simple. A very limited number of cases call for past or future simple tenses and even less times you will be pressed to use more complex tenses.

ikoevska: Here comes in place -> here comes is sufficient

bindings.md
```diff
@@ -137,22 +172,92 @@ source.set("onTap", function(eventData) {
...
+for(i = 0; i < 5; i++) {
+	list.push(new Date());
+}
+source.set("items", list);
+```
+###Binding to a Parent Binding Context
+
+Another common case in working with bindings is requesting access to the parent binding context. The problem comes with dynamically created elements (items) based on some data source. For example `ListView` creates its child items based on an item template, which describes how ListView element will look like. When this element is added to the visual tree, it gets for binding context an element from ListView `items` array (with the corresponding index). This process actually creates a new binding context chain for the child item and its inner UI elements. So no inner UI element cannot access binding context of the `ListView`. In order to solve this problem NativeScript binding infrastructure has two special keywords `$parent` and `$parents`. While the first one denotes the binding context of the direct parent visual element, the second one can be used as an array (with number or string index). This gives the option to choose either `N` levels of UI nesting or get a parent UI element with a given type. Lets see how this works in a real case example.
```
ikoevska: If you're striving for a beginner article - explain what is parent binding context.

ikoevska: The problem comes... suggests that you have defined a problem somewhere.

ikoevska: Comma after "for example".

ikoevska: Very often you can replace the very formal "in order to" with "to". You don't need to do it here but keep it in mind for future writing.

ikoevska: which describes how ListView element will look like -> there's something missing here. You probably meant "how the ListView element will look like" or "how ListView elements will look like".

ikoevska: More often than not you don't need to use actually in your writing.

ikoevska: Avoid beginning sentences with "So". When you do, in most cases, you'll need a comma after it.

ikoevska: In your case, let's is a contraction of let + us and should be spelled "Let's".

bindings.md
```diff
@@ -137,22 +172,92 @@ source.set("onTap", function(eventData) {
...
+
+``` XML
+<Page loaded="pageLoaded">
```
ikoevska, line 215: Please, add code comments in the places that illustrate your lengthy description above. And below.

bindings.md
```diff
@@ -137,22 +172,92 @@ source.set("onTap", function(eventData) {
...
+    viewModel.set("items", [1, 2, 3]);
+    viewModel.set("test", "Test for parent binding!");
+    page.bindingContext = viewModel;
+}
+```
 
-> Note: Be aware that if there is an event handler function **onTap** within the page code behind ([more info about xml declarations](./ui-with-xml.md)), and **onTap** function within the **bindingContext** object then there will be 2 event handlers hooked for that button and both will be executed on tap event.
+The previous example shows how to access a parent binding context of the first `ListView` element. Since the binding context of the Page, GridLayout and ListView is the same (inherited from Page), replacing `$parents['ListView']` with `$parents['GridLayout']` or `$parents['Page']` will not change the result. Also the same result will be achieved with `$parents[1].test` instead of `$parents['ListView'].test`. However string index syntax is better. It will work even when a new UI hierarchy level is introduced (e.g. by surrounding TextField in a another panel). The other keyword `$parent` is actually just a short-cut for `$parents[0]` and will return the direct UI predecessor's binding context.
```
ikoevska: Do you really need to outline this alternative here if your initial example outlines the better practice? When creating technical documentation, you are not supposed to list all the ways you can do something but rather the ones considered a best practice or most straightforward.

bindings.md
```diff
@@ -137,22 +172,92 @@ source.set("onTap", function(eventData) {
...
 
-##Using expressions for bindings
+##Using Expressions for Bindings
 
-A great way of using bindings is the option to create a custom expression (that will be evaluated every time when the source property is changed). Custom expression could help in cases when a certain logic should be applied to the UI while keeping the underlying business data and logic clear. To be more clear lets see a basic binding expression example.
+A great way of using bindings is the option to create a custom expression. Custom expressions could help in cases when a certain logic should be applied to the UI, while keeping the underlying business data and logic clear. To be more clear lets see a basic binding expression example. The result should be a TextField element that will display the value of the `sourceProperty` followed by " some static text" string.
```
ikoevska: - The first sentence doesn't make much sense.
- Two instances of "clear" very close to each other. 
- Incorrect lets spelling.

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+The full binding syntax contains three parameters - first parameter is the source property, which will be listened to for changes, second parameter is the expression that will be evaluated and the third parameter states if the binding is `twoWay` or not. As mentioned earlier by default XML declaration creates a `twoWay` binding by default, so in our example the third parameter could be omitted. We can also omit the first parameter, but this will mean that the expression will be evaluted everytime we have a change in the bindingContext and not only when the property, we are interested in, has changed. Thus, the recommended syntax is to include two parameters in the XML declaration, as in our example - the property of interest and the expression, which has to be evaluated.
```
ikoevska: - the first, the second
- Two instances of by default in the same sentence
- Also, I am confused. Wasn't oneWay the default setting?
- I encourage rewriting this paragraph to simplify it and kill the passive voice.

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
 
-##Using converters in bindings
 
-Speaking of a `twoWay` binding there is a common problem that origins in the different way of storing and displaying data. Probably the best example here is the date and time objects. Generally date and time information is stored as a number or a sequence of numbers (very useful for indexing, searching and other database operations), but this is not the best possible option for displaying date to the application user. Also there is another problem when user inputs a date (in our example user types into a TextField). The result of user input will be a string (TextField.text property) - representation of a date in a convenient to user's culture way usually using the same format as for display `DD.MM.YYYY`. This string should be converted to a correct date object. Lets see how this could be handled with NativeScript binding.
+###Supported Expressions
+NativeScript binding infrastructure supports different kind of expressions including:
```
ikoevska: The NativeScript binding infrastructure...

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+
+| Feature | Example | Description |
+|:--------|:--------|:------------|
+| property access | `obj1.obj2.prop1` | The result will be the value of the `prop1` property of the object `obj2`. Despite the fact expressions in binding are based on `polymer expressions`, which supports re-evaluation of expression when any value within the dot (.) chain is changed. NativeScript uses expressions only in context of bindings (for now), so binding expression will be re-evaluated only when binding's `sourceProperty` is changed (due to performance considerations). Expression part will not observe and therefore will not initiate re-evaluation. |
```
ikoevska: - Despite the fact expressions in binding are based on `polymer expressions`, which supports re-evaluation of expression when any value within the dot (.) chain is changed. -> This is incomplete.
- binding's `sourceProperty` -> the binding `sourceProperty` or the `sourceProperty` of the binding. Unless the binding is a person, don't form possessives for it.
- The expression part

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+| array access | `arrayVar[indexVar]` | Where arrayVar is an array and indexVar is a valid index for that array. |
```
ikoevska: I don't really see a description here.

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+| logical operators | `!var1` | Logical not. |
```
ikoevska: Aim for parallelism in the descriptions.

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+| unary operators | `+var1`, `-var2` | Converts var1 into a number. Converts var2 to number and negates it. |
+| binary operators | `var1 + var2` | Supported operators - `+, -, *, /, %`. |
```
ikoevska: Consider using a colon instead of a dash.

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+| comparison operators | `var1 > var2` | Supported operators - `<, >, <=, >=, ==, !=, ===, !==`. |
+| logical comparison operators | `var1 && var2 || var3`. | Supported operators - `&&, ||`. |
+| ternary operator | `var1 ? var2 : var3` | Evaluates value of `var1` and if truly returns `var2` else `var3` |
```
ikoevska: - the value of var 1
- and if true

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+| grouping parenthesis | `(a + b) * (c + d)` | |
+| function calls | `myFunc(var1, var2, ..., varN)`| Where myFunc is a function available in binding context (used as context for expression) or within `application level resources` and value of the `var1` and `varN` will be used as parameter(s). |
+| filters | `expression | filter1(param1, ...) | filter 2` | Filters actually is an object or a function that is applied to the value of the expression. Within the context of binding this feature is used as converters. For more information see dedicated topic for converters.|
```
ikoevska: Again, link to the article you're referencing to.

bindings.md
```diff
@@ -162,11 +267,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+
+
+##Using Converters in Bindings
+
+Speaking of a `twoWay` binding there is a common problem that origins in the different way of storing and displaying data. Probably the best example here is the date and time objects. Generally date and time information is stored as a number or a sequence of numbers (very useful for indexing, searching and other database operations), but this is not the best possible option for displaying date to the application user. Also there is another problem when the user inputs a date (in our example user types into a TextField). The result of the user input will be a string (TextField.text property) - representation of a date in a convenient to user's culture way. This string should be converted to a correct date object. Lets see how this could be handled with NativeScript binding.
```
ikoevska: - originates (origins is the plural form of the noun)
- there is a common problem - what is this common problem?
- The result of the user input will be a string (TextField.text property) - a culture-specific representation of a date. (Or a similar rewrite - right now the original sentence doesn't make much sentence in English.)

bindings.md
```diff
@@ -230,7 +354,7 @@ source.set("testDate", new Date());
...
+Note the special operator (|) within the expression. The above code snippet (both XML and JavaScript part) will display a date in a `DD.MM.YYYY` format (`toView` function), and when a new date is entered with the same format, it is converted to a valid `Date` object (`toModel` function). `Converter` object should have one or two functions (`toView` and `toModel`) executed every time when a data should be converted. `toView` function is called when data will be displayed to the end user as value of any UI view, and `toModel` function will be called when we have an editable element (like TextField) and the user enters a new value. In case of one way binding `Converter` object could have only `toView` function or it should be a function. All converter functions have an array of parameters, where the first parameter is the value, which will be converted and all other parameters are custom parameters defined in the converter definition.
```
ikoevska: - (both XML and JavaScript part) -> (both XML and JavaScript)
- Go through this paragraph and place the missing articles (the, a, an)

bindings.md
```diff
@@ -254,7 +378,7 @@ source.set("dateFormat", "DD.MM.YYYY");
...
+Setting a converter function and parameter within the bindingContext is very useful, however this is not the case when `listview` items should be bound. The problem comes from the fact that the bindingContext of a `listview` item is a data item, which is a part of any collection (array), and to apply a converter - the converter and its parameters should be added to the data item, which will result in a multiplied converter instances. Tackling this problem with NativeScript is fairly simplified. Binding infrastructure seeks for an application level resources to find a proper converter and parameters. So the converters could be added in the resources in the application modul. To be more clear examine following example (both XML and JavaScript):
```
ikoevska: - very useful for what, in what cases
- "a multiplied converter instances" - there is something wrong with the usage of singular and plural here
- fairly simplified or fairly simple?
- Some the/a/an articles are missing here
- application modul should probably application module

ikoevska:
All in all, you're off to a great start rewriting the Data Binding docs. Kudos! A few things I've noticed and to which I'd like to draw your attention (as general rules for writing documentation):
- Use active voice. You're using mostly passive voice with some could/should overuse. 
- Place your articles (the/a/n).
- Use present simple. It is more than sufficient for most of your writing.
- Link to topics which you are referencing as "the next article", "a dedicated chapter" and so on. Name them and provide links to them (even if it is about a section in the current article).
- Aim to transform your thorough descriptions into code comments. Your dev readers are more likely to pay attention to the code comments, especially when they are appropriately placed.
- Don't go out of your way to explain one thing from multiple angles or to provide more than one solutions to one problem. Go with the best practice or the most straightforward way.
- Limit the instances of "contrary", "however", "actually", "in order to".
- Use complete, simple and short sentences.
- Avoid ambiguity.
- Strive for parallelism in lists and tables. Use similar structures for bullet points and for the text in the same column of a table.
- Aim to break down lengthy paragraphs and sentences.

A suggestion for improving the content here (and throughout actually). Because the examples here seem interconnected, consider having a place in the beginning of each article which explains the context of the code samples that illustrate your concept. This will set the theme for the reader and will let him/her tune into the goal of the current article.

And something for future discussion - we need to figure out if the NativeScript documentation should go into the we/our direction instead of the you/your direction that is the industry standard.

bindings.md
```diff
@@ -162,11 +266,30 @@ A great way of using bindings is the option to create a custom expression (that
...
+##Using Converters in Bindings
+
+Speaking of a two-way binding there is a common problem - having different ways of storing and displaying data. Probably the best example here is the date and time objects. Date and time information is stored as a number or a sequence of numbers (very useful for indexing, searching and other database operations), but this is not the best possible option for displaying date to the application user. Also there is another problem when the user inputs a date (in the example below the user types into a TextField). The result of the user input will be a string, which will be formatted in accordance with the user's ptreferences. This string should be converted to a correct date object. Lets see how this could be handled with NativeScript binding.
```
nsndeck: ptreferences should be preferences

PR #170, opened 2015-10-28
Mention that the Android SDK installation folder may be different.

dimodi:
The installation folder is C:\Users\dimov\AppData\Local\Android\android-sdk if installed via chocolatey, so it might be a good idea to change the default example PATH.


enchev:
Thank you @dimodi!

PR #175, opened 2015-11-06
Changes in the Event article:merged short and full syntax of adding e…

N3ll:
Changes in the Event article:
-merged short and full syntax of adding event listener
-merged short and full syntax of removing listner
-fixed example of how to write getters and setters of Observable


events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
...
-In general, events are very similar to a radio station: the **sender** plays music (fires an event) and the **listener** dances (executes an action).
+Events are very similar to a radio station: the **sender** plays music (fires an event) and the **listener** dances (executes an action).
 
-To repeat, NativeScript streamlines handling events by providing a class called `Observable`. Because it is one of the base classes within the NativeScript framework, almost every NativeScript object (component) has an option for dealing with events.
+NativeScript streamlines handling events by providing a class called `Observable`. Because it is one of the base classes within the NativeScript framework, almost every NativeScript object (component) has an option for dealing with events.
 
-## Adding an Event Listener
+## Adding an Event Handler
 
 Adding an **event listener** means setting a function (method) that executes when the event is raised.
```
ErjanGavalji: `handler` instead of `listener` in the bolded text?

events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
 
 ## Overview
 
-An event is a message sent from an event emitter to signify the occurrence of a specific action. This action can be generated by a user action (such as a tap) or by program logic (for instance to indicate that downloading an image from a server has completed). The object that raises the event is called **event sender** or simply **sender**. The object that consumes the event is called **event listener** or simply **listener**.
+An event is a message sent from an event emitter to signify the occurrence of a specific action. This action can be generated by a user action (such as a tap) or by program logic (for instance to indicate that downloading an image from a server has completed). The object that raises the event is called **event sender** (simply **sender**) or **event handler**. The object that consumes the event is called **event listener** or simply **listener**.
```
ErjanGavalji: `The object that raises the event is called **event sender** (simply **sender**) or **event handler**`
is this correct?

Shall it not be
`The object that raises the event is called **event sender** (simply **sender**) or **event raiser**`
instead?

events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
...
 
-Being a TypeScript (JavaScript) framework, NativeScript cannot benefit from a built in the language event handling mechanism. But because of the events' critical importance for modern mobile application development, NativeScript provides a class `Observable` that powers the process of working with events. Find more information about it in the [API Reference](./ApiReference/data/observable/observable.md).
+Being a TypeScript (JavaScript) framework, NativeScript cannot benefit from a built in the language event-handling mechanism. But because of the events critical importance for the mobile application development, NativeScript provides a class `Observable` that powers the process of working with events. Find more information about it in the [API Reference](./ApiReference/data/observable/observable.md).
```
ErjanGavalji: I'd reword the sentence of `Being a TypeScript (JavaScript) framework, NativeScript cannot benefit from a built in the language event-handling mechanism.` to 'Being a TypeScript (JavaScript) framework, NativeScript cannot benefit from an event-handling mechanism built in the language.`

events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
...
 Adding an **event listener** means setting a function (method) that executes when the event is raised.
 
-The examples below show how to set a function that prints a "Hello World!" message in the console to be executed when a button is tapped. You can choose between the shorthand syntax and the full syntax or you can declare the event handler in XML.
+The examples below shows how to set a function that prints a "Hello World!" message in the console when a button is tapped. You can choose between the shorthand syntax and the full syntax or you can declare the event handler in XML.
 
-### Adding an Event Handler Using the Shorthand Syntax
+### Adding an Event Handler Using the Shorthand or Full Syntax
 
-Use the shorthand syntax to write an in-line function to be executed on an event (tapping a `testButton` in this example):
+In the following code, there is an example of how to add an event listener by using the short and full syntax. There is a third iptional parameter that represents the `this` argument.
```
ErjanGavalji: `handler` or `listener`? :)
and a small typo - `optional` instead of `iptional`

events.md
```diff
@@ -115,36 +108,31 @@ export function onTap(eventData) {
 
 ## Removing an Event Listener
 
-Under most circumstances you don't need to remove the event listener. If you do, however, for example if you want to receive the event just once or want to remove the event listener to free up resources, you can resort to the methods described in the sections below.
+Under most circumstances you don't need to remove the event listener. If you do, for example if you want to receive the event just once or want to remove the event listener to free up resources, you can resort to the methods described in the sections below.
```
ErjanGavalji: I'd put the dependent clause in parentheses to improve the readability:

`If you do (for example when you want to receive the event just once or want to remove the event listener to free up resources), you can resort to the methods described in the sections below.`

events.md
```diff
@@ -115,36 +108,31 @@ export function onTap(eventData) {
...
 
 > There is no syntax to remove an event listener through an XML declaration.
 
-### Removing an Event Listener Using the Shorthand Syntax
+### Removing an Event Listener Using the Shorthand or Full Syntax
 
-The following example uses the shorthand syntax to remove all listeners for the tap event of the testButton instance. If there are more than one event listener objects, you can set a second parameter with the name of the callback function. This way only the referenced event listener is removed.
+The following example uses the shorthand and full syntax to remove all listeners for the tap event of the testButton instance. If there are more than one event listener objects, you can set a second parameter with the name of the callback function. This way only the referenced event listener is removed. There is also a third optional parameter, which is used when multiple event listeners with different `this` arguments are available.
```
ErjanGavalji: `syntaxes` instead of `syntax`?

events.md
ikoevska: Comma after "for instance", "for example", etc.

events.md
ikoevska: :+1: on Erji's comment. This sentence hardly makes sense.

Even with the rephrasing, I honestly cannot see any value to this introductory sentence.  Probably rework the whole paragraph into something like "The NativeScript framework provides an event-handling mechanism via the `Observable` class." Everything else is quite unnecessary in my opinion.

events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
...
 
-Being a TypeScript (JavaScript) framework, NativeScript cannot benefit from a built in the language event handling mechanism. But because of the events' critical importance for modern mobile application development, NativeScript provides a class `Observable` that powers the process of working with events. Find more information about it in the [API Reference](./ApiReference/data/observable/observable.md).
+Being a TypeScript (JavaScript) framework, NativeScript cannot benefit from a built in the language event-handling mechanism. But because of the events critical importance for the mobile application development, NativeScript provides a class `Observable` that powers the process of working with events. Find more information about it in the [API Reference](./ApiReference/data/observable/observable.md).
 
 ## How Do Events Work
 
-In general, events are very similar to a radio station: the **sender** plays music (fires an event) and the **listener** dances (executes an action).
+Events are very similar to a radio station: the **sender** plays music (fires an event) and the **listener** dances (executes an action).
```
ikoevska: Blog-like phrasing. While it's quirky and cute, it is not documentation style.

events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
...
 
-To repeat, NativeScript streamlines handling events by providing a class called `Observable`. Because it is one of the base classes within the NativeScript framework, almost every NativeScript object (component) has an option for dealing with events.
+NativeScript streamlines handling events by providing a class called `Observable`. Because it is one of the base classes within the NativeScript framework, almost every NativeScript object (component) has an option for dealing with events.
```
ikoevska: I sense redundancy in the force.

You have the same statement a few lines above. Provide this information in just one place.

events.md
ikoevska: You might want to go with "To add an event listener is to set a function..."

events.md
```diff
@@ -21,73 +21,66 @@ This article contents:
...
 
-## Adding an Event Listener
+## Adding an Event Handler
 
 Adding an **event listener** means setting a function (method) that executes when the event is raised.
 
-The examples below show how to set a function that prints a "Hello World!" message in the console to be executed when a button is tapped. You can choose between the shorthand syntax and the full syntax or you can declare the event handler in XML.
+The examples below shows how to set a function that prints a "Hello World!" message in the console when a button is tapped. You can choose between the shorthand syntax and the full syntax or you can declare the event handler in XML.
```
ikoevska: examples ... shows

Align the numbers =)

events.md
ikoevska: iptional -> optional

@ErjanGavalji, way above there is a clear definition which states that listener and handler are synonyms. Using them interchangeably is perfectly acceptable and expected to avoid the repetition of the same term over and over again.

events.md
```diff
@@ -99,7 +92,7 @@ Another option to set an event handler is to use an XML declaration like this:
   </StackLayout>
 </Page>
 ```
-Naturally, you need a code behind file to write the function body (the code behind file has the same file name but a different extension: .js or .ts depending on the language you are using). It could look like this:
+You need a code-behind file to write the function body (the code-behind file has the same file name, but a different extension: .js or .ts depending on the language you are using). It could look like this:
```
ikoevska, line 88: Now this is precisely why you need a Glossary doc =) In this case, you could have simply linked to the definition rather than explaining it in brackets.

events.md
ikoevska: "Under most circumstances" could be simplified - often, usually, typically, etc. are viable synonym candidates. I am saying this because when you write documentation, you need to think of the content localization possibility (or the possibility of running your text through machine translators). You need to keep it short and simple - the less pompous phrasing and the shorter phrasing - the better.

I'd go with the following phrasing here:
"Usually, you don't need to remove the event listener. You might need to do it when you want to receive the event just once or to free up resources. In such cases, you can apply the following methods."

Most of the time you don't need to say "described below", "outlined in the next lines", "explained in this chapter", etc. These are pretty much self-evident.

events.md
ikoevska: Would you use syntaxes if this sentence was in Bulgarian ;) I'd be interested to hear you say it out loud.

Please, reduce the "there is/there are" phrasing. I am a firm believer that this phrasing does not bring anything of value to the table - just empty words.

events.md
```diff
@@ -115,36 +108,31 @@ export function onTap(eventData) {
...
+testButton2.removeEventListener(buttonModule.Button.tapEvent);
 ```
-``` TypeScript
-testButton.removeEventListener(buttonModule.Button.tapEvent, onTap);
-```
-
 ## PropertyChange Event
 
-The `Observable` class provides a built-in event called `propertyChange` which is called when a property is changed. This is how to subscribe to this event:
+The `Observable` class provides a built-in event called `propertyChange`, which is called when a property is changed. This is how to subscribe to this event:
```
ikoevska: The punctuation rules in English do not follow our own punctuation rules. 

You don't always need the comma before a "which/who/that" part of the sentence. The rule of thumb is that you place the comma if you want to emphasize that this part of the sentence could be or could not be read by the user - based on their discretion. The comma usually indicates that this part is not so important and could safely be omitted. If you do not place the comma, however, you're saying "this must be read because it is vital to the part before it".

events.md
N3ll, line 88: This is considered to be added in future revisions of the documentation :)

ErjanGavalji:
:+1:

ikoevska:
You can go ahead and release this as it is. However, you can run a check for "there is/there are" (I saw a few instances of those) and you can review the usage of passive voice in a few spots.

Generally, passive voice should be used sparsely and is acceptable in the following cases:
- When the subject performing the action is unknown.
- When the subject performing the action is of little or no importance.
- When the action is more important than the subject performing it.

In your case, the usage of passive voice mostly follows these rules.

PR #182, opened 2015-11-13
New System requirements and setup instructions

teobugslayer:


teobugslayer:
Ping @ikoevska & @N3ll

N3ll:
:+1:

ErjanGavalji:
:+1:

ErjanGavalji:
Do you think we should continue claiming the support for node 0.10.x? It's archaic...

teobugslayer:
@ErjanGavalji , according the Analytics stats, almost 19% of users are using node.js 0.10. If we are going to obsolete it, we should warn the users in advance. Perhaps the 1.5 release is a good opportunity to do so.

ErjanGavalji:
@teobugslayer Well, 19% seem a lot to me (to break I mean)... Are we sure we ONLY count customers (not our build/test machines) having that installed on their machines? What is the timespan of these stats?

ikoevska:
:+1:

PR #185, opened 2015-11-23
iOS plugins fix

ivanbuhov:
- Makes it clear that `platforms/ios` folder should not contains ordinary libraries (e.g. `.dylib` files) but only frameworks (`.framework`) and also the frameworks can be both static and shared.
- Fix a typo.


jasssonpet:
:+1:

PR #205, opened 2015-12-18
added common document describing common debugging commands and options

Plamen5kov:
related to: https://github.com/NativeScript/android-runtime/issues/290

### DIFF
--- a/core-concepts/debugging.md
+++ b/core-concepts/debugging.md
@@ -8,36 +8,44 @@ previous_url: /debugging
 
 # Debugging
 
-This article contents:
+You can debug apps developed with the NativeScript framework from the NativeScript CLI.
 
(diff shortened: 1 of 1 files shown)


ikoevska:
I've applied some changes from editing. However, this document is a bit... bland to my taste and mainly quotes the CLI help. I'd like to see more information on the built-in tools. For example, what are they based on, are there any prerequisites for debugging, some overview of the built-in tools and how to work with them, etc.

KristinaKoeva:
I am going to merge this one because otherwise we won't have any doc page that describes the debugging in general. Furthermore we need to reference it in our more detailed debugging pages like this one http://docs.nativescript.org/runtimes/ios/how-to/AppInspector . 
We are going to extend the content of this document in some other iteration.

PR #230, opened 2016-02-01
fixed bug that broke navigation after login failed.

rz-robsn:
Hi,

I was trying out the "Getting Started" guide when I encountered a small bug in part 4.1. This pull request fixes it.

The steps to reproduce are : 
- paste the code snippet described in part 4.1 in the project (after following all the previous steps, obviously)
- launch the app
- enter account credentials that don't exist 
- the error popup appears.
- then enter credentials that do exists
- => nothing happens when I expected to navigate to another page.

I reproduced this error on the ios simulator 8.4. The console also logged the following output :
`sampleGroceries[21575]: pushViewController:animated: called on <UINavigationControllerImpl 0x7b967a60> while an existing transition or presentation is occurring; the navigation stack will not be updated.
`

That's it. Thank you for a great guide so far !

### DIFF
--- a/start/getting-started/chapter-4.md
+++ b/start/getting-started/chapter-4.md
@@ -155,16 +155,16 @@ Next, re-write your `signIn()` function to look like this:
 ``` JavaScript
 exports.signIn = function() {
     user.login()
+        .then(function() {
+            frameModule.topmost().navigate("views/list/list");
+        })
         .catch(function(error) {
             console.log(error);
             dialogsModule.alert({
                 message: "Unfortunately we could not find your account.",
                 okButtonText: "OK"
             });
-        })
-        .then(function() {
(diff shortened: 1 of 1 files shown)


ikoevska:
Hey @tjvantoll, do you mind taking a look at this one?

tjvantoll:
Hey @rz-robsn,

Thanks a lot for reporting this! I recently made a change to put the `catch()` handler first, because when the `then()` comes first, it swallows any others that occur during the navigation. So if you had malformed XML on the list page for instance, the login page’s `catch()` handler would unintuitively get invoked and debugging was painful.

However, when I made that switch I forgot one very important line, which I just added in https://github.com/NativeScript/docs/commit/5781f9361e53bad24ee796303523672b89513dcc. Thanks again for pointing this out!

PR #231, opened 2016-02-01
Added missing Promise.reject

FrancoisCamus:
Hi,

In order to get the behavior described in the tutorial:

"This code handles both a successful and unsuccessful login. On success, you call the frame module's `navigate()` method to navigate the user to the (currently empty) list page. On failure, you use the dialog module to show the user an error message. Try inputting some invalid credentials to see what the dialog looks like."

You need to add `return Promise.reject();` in the end of the user.login().catch function. Otherwise the user would still be redirected to the list on login failure.

I looked at the repo for the Groceries app and the code is there.

Great tutorial by the way :)

### DIFF
--- a/start/getting-started/chapter-4.md
+++ b/start/getting-started/chapter-4.md
@@ -161,6 +161,7 @@ exports.signIn = function() {
                 message: "Unfortunately we could not find your account.",
                 okButtonText: "OK"
             });
+            return Promise.reject();
         })
         .then(function() {
             frameModule.topmost().navigate("views/list/list");


tjvantoll:
Thanks @FrancoisCamus! Crazily enough this was also reported in #230 so [I patched it up earlier today](https://github.com/NativeScript/docs/commit/5781f9361e53bad24ee796303523672b89513dcc) before even noticing this. Thanks for reporting it though! If you have any other feedback about the tutorial let me know.

PR #232, opened 2016-02-01
Add note about iOS images

mmillican:


ErjanGavalji:
Well, how about JPEG, TIFF, GIF?

http://stackoverflow.com/questions/3404530/what-are-the-supported-image-file-formats-for-display-on-the-iphone

ikoevska:
Something seems very wrong with this PR. Apart from the single note it contains from @mmillican it also contains multiple commits from other people which, I believe, are already live. I am guessing that some rebasing must be done or that the production branch must be updated? @ErjanGavalji would you please check the state of the production branch and why is this PR in such a state?

As for the comment about the supported formats, it seems that [PNG is the recommendation for app icons](https://developer.apple.com/library/ios/documentation/UserExperience/Conceptual/MobileHIG/IconMatrix.html) (not for any and all images loaded in the app which is the topic discussed by the Stack Overflow thread quoted above). I would change the note to:

```
**IMPORTANT:** The required format for iOS icons is non-interlaced PNG.
```

ErjanGavalji:
@ikoevska Yes, it has been developed on a branch while the tip of the target branch has moved and has been reset (we reset the tip of the production branch each release). The merge is not a problem, we will do it internally if @mmillican can't find the resources to.

ErjanGavalji:
Done.
Closing this one.

PR #233, opened 2016-02-05
Add `text-decoration` property to styling

burkeholland:
Looks like there is already support for `line-through` and `underline` with `text-decoration`. Adding to the docs.


tjvantoll:
By chance I just happened to run into this as well 😀 There’s also `"none"`.

PR #234, opened 2016-02-09
Add ANDROID_HOME to Path

burkeholland:
I had to add the ANDROID_HOME variable to my path in order to get it to work.


teobugslayer:
:-1: this is a bug, or at least something we must investigate.

ErjanGavalji:
@teobugslayer, let's close the PR if we have created an issue for the investigation?

teobugslayer:
@burkeholland , can you create a issue in the CLI repository with the output produced by `tns doctor --log trace`?

PR #248, opened 2016-02-26
Update ns-setup-os-x.md

Lampei:
Added notes about setting environment variables if they do not get set correctly using the terminal window.  Updated the name of the support libraries.


teobugslayer:
Hello @Lampei,

It looks like this PR breaks the formatting of the article. I suggest to add one empty before and after the code blocks. You can approximately verify the final look if you click on "View" button in the PR - this will render the markdown.

teobugslayer:
Thank you!

PR #256, opened 2016-03-06
Added notes about installing cocoapods

Lampei:
I ran into this issue and found the above solution in the NativeScript google group, so I thought I would add a short note about how to fix the problem in the troubleshooting section.

### DIFF
--- a/start/troubleshooting.md
+++ b/start/troubleshooting.md
@@ -52,6 +52,20 @@ A problem occurred configuring root project 'app_name'.
 
 **Workaround:** Update the opener package to a newer version. If this does not resolve the issue, try modifying your local copy of the package.
 
+### Cocoapod install fails on OS X system
+
+**Problem** The cocoapod install fails when creating a shallow clone of spec repo 'master'.  The actual error should look something like
+```
+Creating shallow clone of spec repo 'master' from 'https://github.com/CocoaPods/Specs.git'
+[!] Unable to add a source with url 'https://github.com/CocoaPods/Specs.git' named 'master'.
+You can try adding it manually in '~/.cocoapods/repos' or via 'pod repo add'.
+```
+
+**Workaround** There are 3 methods to try to fix this issue:
+  1. Make sure that you tried to install cocoapods initially by `sudo gem install cocoapods`
+  1. Type `pod setup`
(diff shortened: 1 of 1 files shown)


start/troubleshooting.md
```diff
@@ -52,6 +52,20 @@ A problem occurred configuring root project 'app_name'.
 
 **Workaround:** Update the opener package to a newer version. If this does not resolve the issue, try modifying your local copy of the package.
 
+### Cocoapod install fails on OS X system
```
ikoevska, line 55: CocoaPods installation fails on OS X systems

start/troubleshooting.md
```diff
@@ -52,6 +52,20 @@ A problem occurred configuring root project 'app_name'.
...
+
+**Problem** The cocoapod install fails when creating a shallow clone of spec repo 'master'.  The actual error should look something like
```
ikoevska, line 57:
```
**Problem:** The CocoaPods installation fails on an OS X system with the following error message: 
```

start/troubleshooting.md
```diff
@@ -52,6 +52,20 @@ A problem occurred configuring root project 'app_name'.
...
+```
+Creating shallow clone of spec repo 'master' from 'https://github.com/CocoaPods/Specs.git'
+[!] Unable to add a source with url 'https://github.com/CocoaPods/Specs.git' named 'master'.
+You can try adding it manually in '~/.cocoapods/repos' or via 'pod repo add'.
+```
+
+**Workaround** There are 3 methods to try to fix this issue:
```
ikoevska, line 64: Since these are three separate approaches (and not three consecutive steps), it is a good idea to go with an unnumbered list.

```
**Workaround:** There are three methods to try to fix this issue:
* Make sure that you tried to install CocoaPods initially by `sudo gem install cocoapods`
* Type `pod setup`
* If neither of the above fixes the problem, try manually adding the directories `~\.cocoapods` and `~\.cocoapods\repos`
```

ErjanGavalji:
@Lampei Thanks a ton for the contribution!
Can you address @ikoevska's notes?

I'll take care of the merge afterwards...

ErjanGavalji:
:+1: after merging @ikoevska's notes

Lampei:
Done and new pull request created.

PR #261, opened 2016-03-09
Logically separate installation of the CLi from setting up the depend…

teobugslayer:
…encies

### DIFF
--- a/start/quick-setup.md
+++ b/start/quick-setup.md
@@ -27,6 +27,15 @@ To take advantage of this autonomy, you need to configure your system with the t
 
 ### Quick setup
 
+NativeScript CLI is a Node.js application. If you have Node.js already installed, just run the following command.
+
+```Shell
+npm i -g nativescript
+```
+
+If you do not have Node.js, you need to install it first. Here's how to install it on [Mac]({% slug osx %}), [Windows]({% slug windows %}) or [Linux]({% slug linux %})
+
+To develop your apps, you need the Native SDK dependencies.
 If this is your first time developing a mobile project, consider using the one-liner scripts in this section to effortlessly setup your machine. 
 If you have experience developing mobile apps, you may skip to the [advanced setup](#advanced-setup) section below.
 


start/quick-setup.md
```diff
@@ -27,6 +27,15 @@ To take advantage of this autonomy, you need to configure your system with the t
 
 ### Quick setup
 
+NativeScript CLI is a Node.js application. If you have Node.js already installed, simply run the following command.
+
+    ```Shell
+    npm i -g nativescript
+    ```
+
+If you do not have Node.js, you need to install it first. Refer to our [Mac]({% slug osx %}) and [Windows]({% slug windows %}) guides for instructions.
```
rosen-vladimirov: What about Linux?

start/quick-setup.md
```diff
@@ -27,6 +27,15 @@ To take advantage of this autonomy, you need to configure your system with the t
 
 ### Quick setup
 
+NativeScript CLI is a Node.js application. If you have Node.js already installed, simply run the following command.
```
ikoevska: The NativeScript CLI

just run

start/quick-setup.md
```diff
@@ -27,6 +27,15 @@ To take advantage of this autonomy, you need to configure your system with the t
...
+
+    ```Shell
```
ikoevska: No need for indentation in this block. In fact, it might cause issues in the output.

start/quick-setup.md
ikoevska: Refer to our [Mac]({% slug osx %}) and [Windows]({% slug windows %}) guides for instructions.

I would probably go with "Here's how to install it on Mac (link), Windows (link) or Linux (link)."

Also, I am not really sure why you refer to our setup guides for Node.js installation. Can't you just say "If you do not have Node.js, you can download and install it from [website](link)."

teobugslayer: Because our guides uses Homebrew. When installing Node.js from the site some commands must be used with "sudo". This conflicts with other parts of the documentation which expects the machine to be set through homebrew.

ikoevska: Oh boy, yeah, I totally forgot that. Well then, go with the first or the second option in my previous comment :) Your choice :)

rosen-vladimirov:
:+1:

PR #263, opened 2016-03-10
Add `returnPress` and `returnKeyType` to TextField How-To

burkeholland:
Title explains it all. I was looking for this yesterday and had to refer to a github project to find it.

### DIFF
--- a/ApiReference/ui/text-field/HOW-TO.md
+++ b/ApiReference/ui/text-field/HOW-TO.md
@@ -87,7 +87,7 @@ model.set("secure", false);
 // textField.secure is now false
 ```
 ### Binding to the return press event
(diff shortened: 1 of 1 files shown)


ApiReference/ui/text-field/HOW-TO.md
```diff
@@ -86,3 +86,15 @@ textField.bind(options, model);
 model.set("secure", false);
 // textField.secure is now false
 ```
+### Binding to the return press event
+It is possible to bind to the return press event on a TextField. This event will be fired with the user hits the "Return" key on the keyboard while in the TextField.
```
ikoevska: will be fired with the user
->

Perhaps, "when the user" instead of "with the user"?

ErjanGavalji:
:+1: after applying @ikoevska's note

PR #283, opened 2016-03-23
Fixed android support repository name

TsvetanMilanov:
Replaced Android Support Repository name with Local Maven repository for Support Libraries.


TsvetanMilanov:
:+1:  from @rosen-vladimirov

PR #286, opened 2016-03-25
Update accessing-native-apis-with-javascript.md

erikruth:
Some minor editing changes and a few comments. 
From the Telerik style guide:
e.g. is commonly used in documentation in place of for example. It always requires a comma after the second period.

### DIFF
--- a/core-concepts/accessing-native-apis-with-javascript.md
+++ b/core-concepts/accessing-native-apis-with-javascript.md
@@ -13,21 +13,21 @@ In this article, you will learn how to call native APIs from JavaScript with var
 
 # Numeric Types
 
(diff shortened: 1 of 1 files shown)


core-concepts/accessing-native-apis-with-javascript.md
```diff
@@ -84,7 +84,7 @@ var str = new java.lang.String('Hello world!');
 var result = str.endsWith('world!');
 console.log(result); // true
 ```
-
+<Comment: My expectation is that I should see the word boolean somewhere in the previous Android example.>
```
ErjanGavalji: the `endsWith` method returns a boolean :)

ErjanGavalji:
:+1: After removing the comment

PR #287, opened 2016-03-25
Update bindings.md

erikruth:
There are several comments to address in this.


core-concepts/bindings.md
```diff
@@ -443,11 +443,11 @@ appModule.resources["dateConverter"] = dateConverter;
...
 
 ##Stop Binding
 
-Generally there is no need to stop binding explicitly, since Binding object uses weak references, which prevents any memory leaks. However, there are some scenarios, where binding must be stopped. In order to stop existing data binding just call **unbind** method with target property name as argument.
+Generall,y there is no need to stop binding explicitly since a Binding object uses weak references, which prevents any memory leaks. However, there are some scenarios where binding must be stopped. In order to stop an existing data binding, just call the **unbind** method with the target property name as the argument.
```
ikoevska, line 513: A typo perhaps? ->Generally,

Yes. Good catch.

core-concepts/bindings.md
```diff
@@ -37,12 +37,12 @@ Part of the data binding settings is the way data flows. NativeScript data bindi
 
 ##Basic Binding Concepts
 
-Generally almost every UI control could be bound to a data object (all NativeScript controls are created with data binding in mind). After your code has ,et the following requirements, you can use data-binding out of the box.
+Generally, almost every UI control could be bound to a data object (all NativeScript controls are created with data binding in mind). After your code has met the following requirements, you can use data-binding out of the box.
 
 * The target object has to be a successor of the **Bindable** class. All NativeScript UI controls already inherit from this class.
-* For **two-way** data binding, the target property should to be a **dependency property**.
```
ErjanGavalji, line 43: A "dependency property" is a term, denoting a specific type of property, hence "should be a **dependency property** would fit best.

Fixed--ER

core-concepts/bindings.md
```diff
@@ -50,7 +50,7 @@ Generally almost every UI control could be bound to a data object (all NativeScr
 
 The example below consists of a `Label`, `TextField` and a source property to which the UI controls are bound. The purpose will be, when the user enters an input in the `TextField`, to update the property in the code and the `Label` text.
 
-First, the **source** object is created with a **textSource** property. A constant flow of progating changes from the source property to the Label is necessary. Thus the property in the code has to raise a **propertyChange** event, in order to notify the `Label` for the changes. To raise this event, an built-in class is used, which provides this functionality - `Observable`.
+First, the **source** object is created with a **textSource** property. A constant flow of progating changes from the source property to the Label is necessary. <Comment: What is progating? Do you mean propagating? A bit later in the article, five code snippets down, you use propagate.> Thus, the property in the code has to raise a **propertyChange** event in order to notify the `Label` for the changes. To raise this event, a built-in class is used, which provides this functionality - `Observable`.
```
ErjanGavalji, line 60: Yep, it should be _propagating_. With that, "a constant flow of propagating changes" does not make a lot of sense to me. How about "a constant change propagation" instead?

core-concepts/bindings.md
```diff
@@ -83,7 +83,7 @@ var targetTextField = new textFieldModule.TextField();
 import labelModule = require("ui/label");
 var targetLabel = new labelModule.Label();
 ```
-After that the target objects bind to the source object. The TextField uses a two-way binding, so the user input could change the property in the code. And the binding of the Label is set to one-way in order to propagate changes only from the code to the UI.
+After that, the target objects bind to the source object. <Comment: "that" in the previous sentence is vague to me. I suggest you replace "that" with "bound target objects" or whatever "that" refers to.> The TextField uses a two-way binding, so the user input could change the property in the code. And the binding of the Label is set to one-way in order to propagate changes only from the code to the UI.
```
ErjanGavalji, line 97: "After that" is used instead of "as a next step" here to avoid duplication with the "Next," in line 66... Can you suggest something better for sequencing?

I'll change to "Finally,"  I didn't get the sequencing. I was trying to figure out specifically what "that" refers to.

core-concepts/bindings.md
```diff
@@ -122,8 +122,8 @@ targetLabel.bind(labelBindingOptions, source);
...
-
+To create a binding in XML, a source object is needed, which will be created the same way, as in the exmple above ([Two-Way Binding in Code](#two-way-binding-in-code)). Then the binding is described in the XML (using a mustache syntax). With an XML declaration, only the names of the properties are set - for the target: text, and for source: textSource. The interesting thing here is that the source of the binding is not specified explicitly. More about this topic will be discussed in the [Binding source](#binding-source) article.
+<Comment: Since your articles are not numbered in sequence to read them in a particular order, I removed the reference to the "next" article. The customer will read what they want to next.>
```
ErjanGavalji: :+1:

core-concepts/bindings.md
```diff
@@ -133,13 +133,13 @@ To create a binding in XML, a source object is needed, which will be created the
...
+An important part of the data binding is setting the source object. For a continuous flow of data changes, the source property needs to emit a **propertyChange** event. NativeScript data binding works with any object that emits this event. Adding a binding **source** happens by passing it as a second parameter in the method **bind(bindingOptions, source)**. This parameter is optional and could be omited, in which case for source is used a property named **bindingContext** of the `Bindable` class. <Comment: The previous sentence does not make sense to me. "for source is used a property named..." might be "source is used for a property named..."> What is special about this property is that it is inheritable across the visual tree. This means that a UI control can use the `bindingContext` of the first of its **parent** elements, which has an explicitly set **bindingContext**. In the example from [Two-Way Binding in Code](#two-way-binding-in-code), the `bindingContext` can be set either on a `Page` instance or a `StackLayout` instance and the `TextField` will inherit it as a proper source for the binding of its "text" property.
```
ErjanGavalji, line 154: It'd rather be "...in which case, a property of the `Bindable` class, named **bindingContext** is used for the `source` parameter.

core-concepts/bindings.md
```diff
@@ -153,7 +153,7 @@ stackLayout.bindingContext = source;
 ```
 ###Binding to an Event in XML
 
-There is an option to bind a function to execute on a specific event (MVVM command like). This option is available only through an XML declaration. To implement such a functionality, the source object should have an event handler function.
+There is an option to bind a function to execute on a specific event (MVVM command like). <Comment: Do you mean "like a MVVM command"> This option is available only through an XML declaration. To implement such a functionality, the source object should have an event handler function.
```
ErjanGavalji, line 171: :+1:

core-concepts/bindings.md
```diff
@@ -211,7 +211,7 @@ source.set("items", list);
...
...formation, which the child has to use. Generally, the bindingContext is inheritable, but not when the elements (items) are created dynamically based on some data source. For example, `ListView` creates its child items based on an `itemТemplate`, which describes what the `ListView` element will look like. When this element is added to the visual tree, it gets for binding context an element from a ListView `items` array (with the corresponding index). This process creates a new binding context chain for the child item and its inner UI elements. So, no inner UI element cannot access binding context of the `ListView`. <Comment: The previous sentence is odd to me. Do you mean, "So, the inner UI element cannot access the binding conext of the 'ListView'."> In order to solve this problem, NativeScript binding infrastructure has two special keywords: `$parent` and `$parents`. While the first one denotes the binding context of the direct parent visual element, the second one can be used as an array (with a number or string index). This gives you the option to choose either `N` levels of UI nesting or get a parent UI element with a given type. Let's see how this works in a realistic example.
```
ErjanGavalji, line 243: I'd change "it gets for binding context an element from a ListView `items` array (with the corresponding index)" to "for a binding context it gets a corresponding element from the `items` array of the ListView".

:+1:  for the suggestion in the comment

core-concepts/bindings.md
```diff
@@ -276,17 +276,17 @@ NativeScript supports different kind of expressions including:
...
+| unary operators | `+var1`, `-var2` | Converts var1 into a number. Converts var2 to a number and negates it. |
 | binary operators | `var1 + var2` | Adding the value of var2 to var1. Supported operators: `+, -, *, /, %`. |
 | comparison operators | `var1 > var2` | Comparing whether the value of var1 is more than the value of var2. Other supported operators - `<, >, <=, >=, ==, !=, ===, !==`. |
-| logical comparison operators | `var1>1 && var2>1`. | Evealuating whether the value of var1 is more than 1 AND the value of var2 is more than 2. Supported operators: `&&, ||`. |
-| ternary operator | `var1 ? var2 : var3` | Evaluating the value of `var1` and if true returns `var2` else returns `var3` |
-| grouping parenthesis | `(a + b) * (c + d)` | |
+| logical comparison operators | `var1>1 && var2>1`. | Evaluating whether the value of var1 is more than 1 AND the value of var2 is more than 2. Supported operators: `&&, ||`. |
+| ternary operator | `var1 ? var2 : var3` | Evaluating the value of `var1` and if true, returns `var2`, else returns `var3`. |
+| grouping parenthesis | `(a + b) * (c + d)` |<Comment: You are missing description text here.> |
```
ErjanGavalji, line 331: The comment would be something similar to "returns the result of the equation, with the expressions in the parenthesis preceding the multiplication"

core-concepts/bindings.md
```diff
@@ -385,7 +385,7 @@ source.set("dateFormat", "DD.MM.YYYY");
...
+Setting a converter function and a parameter within the bindingContext is very useful for ensuring proper convertion of data. However, this is not the case when `listview` items should be bound. The problem comes from the fact that the bindingContext of a `listview` item is a data item, which is a part of `any` collection (array), and to apply a converter - the converter and its parameters should be added to the data item, which will result in multiple converter instances. Tackling this problem with NativeScript is fairly simple. Binding infrastructure seeks for an application level resources to find a proper converter and parameters. <Comment: The previous sentence does not make sense to me. What is the subject and what action is it doing?> Soyou could add the converters in the resources in the application module. To be more clear, examine the following example (both XML and JavaScript):
```
ErjanGavalji, line 442: Addressing the comment, it would mean something like:

To find proper converter and parameters the binding functionality searches application-level resources.

And, probably a typo: Soyou... -> So you...

ErjanGavalji:
:+1: After addressing the comments...

Erjan, there are still 7-8 code examples that need SEO-friendly captions. Can someone on your team write them please? Also, there are three in-line comments for tech review (look for one after Example 6 and two after Example 7). Thank you -Erik

core-concepts/bindings.md
```diff
@@ -209,10 +211,11 @@ for(i = 0; i < 5; i++) {
...
...formation, which the child has to use. Generally, the bindingContext is inheritable, but not when the elements (items) are created dynamically based on some data source. For example, `ListView` creates its child items based on an `itemТemplate`, which describes what the `ListView` element will look like. When this element is added to the visual tree, it gets for binding context an element from a ListView `items` array (with the corresponding index). This process creates a new binding context chain for the child item and its inner UI elements. So, no inner UI element cannot access binding context of the `ListView`. <Comment: The previous sentence is odd to me. Do you mean, "So, the inner UI element cannot access the binding conext of the 'ListView'."> In order to solve this problem, NativeScript binding infrastructure has two special keywords: `$parent` and `$parents`. While the first one denotes the binding context of the direct parent visual element, the second one can be used as an array (with a number or string index). This gives you the option to choose either `N` levels of UI nesting or get a parent UI element with a given type. Let's see how this works in a realistic example.
```
erikruth, line 245: Please address the comment in the middle of the previous paragraph.

core-concepts/bindings.md
```diff
@@ -385,8 +390,9 @@ source.set("dateFormat", "DD.MM.YYYY");
...
+Setting a converter function and a parameter within the bindingContext is very useful for ensuring proper convertion of data. However, this is not the case when `listview` items should be bound. The problem comes from the fact that the bindingContext of a `listview` item is a data item, which is a part of `any` collection (array), and to apply a converter &mdash; the converter and its parameters should be added to the data item, which will result in multiple converter instances. Tackling this problem with NativeScript is fairly simple. Binding infrastructure seeks for an application level resources to find a proper converter and parameters. <Comment: The previous sentence does not make sense to me. What is the subject and what action is it doing?> Soyou could add the converters in the resources in the application module. To be more clear, examine the following example (both XML and JavaScript):
```
erikruth, line 444: Please address the comment in the previous paragraph.

core-concepts/bindings.md
ErjanGavalji, line 245: Your suggestion is the correct one. I'd even reword it as:

"Thus, the inner UI element cannot access..."

PR #289, opened 2016-03-28
Many comments to address

erikruth:
Since you've agreed with my suggestion to add SEO-friendly code captions and figure captions, I thought I'd start with this article which has many good examples that I think a customer would use Google search to try an specifically find, plus several examples where after a snippet you refer back to the previous example. I've marked the more important (longer) code samples and not snippets that are less than 5 lines. Use your own judgement about whether a code sample is significant enough.
Do think to yourself, "If I were a customer, what would I type into a Google search box to find this code snippet?" and make sure that phrase, in that order if it makes sense, appears in the code or figure caption. This will help you increase findability over the long run.


core-concepts/events.md
```diff
@@ -29,9 +29,9 @@ The NativeScript framework provides a class `Observable` that powers the process
 
 To add an **event handler** means setting a function (method) that executes when the event is raised.
 
-The example below shows how to set a function that prints a "Hello World!" message in the console when a button is tapped. You can choose between the shorthand syntax and the full syntax or you can declare the event handler in XML.
+__Example 1__ shows how to set a function that prints a "Hello World!" message in the console when a button is tapped. You can choose between the shorthand syntax and the full syntax or you can declare the event handler in XML.
 
-### Adding an Event Handler Using the Shorthand or Full Syntax
+###Example 1 (JavaScript): Adding an event handler or event listener using she short and full syntax.
 
 The example below shows how to add an event listener by using the short and full syntax. There is a third optional parameter that represents the `this` argument.
```
ErjanGavalji, line 35: This text is not JavaScript-specific. Let's copy it to the TypeScript example as well.

core-concepts/events.md
```diff
@@ -181,12 +189,13 @@ export class MyClass extends observableModule.Observable {
   }
 }
 ```
-The previous code snippet fires the `propertyChange` event when the property value is changed.
+The code snippet in __Example 5__ fires the `propertyChange` event when the property value is changed.
 
 ## Creating a Custom Event
 
-If your business logic demands it, you may want to fire (raise or emit) a custom event on a particular action. To do that, call the `Observable.notify()` method when the action is completed. This method takes any **implementer** of the [EventData interface]({{site.baseurl}}/ApiReference/data/observable/EventData.md) as event data. It includes basic information about an event&mdash;its name as `eventName` and an instance of the event sender as `object`).
+If your business logic demands it, you may want to fire (raise or emit) a custom event on a particular action (see __Example 6__). To do that, call the `Observable.notify()` method when the action is completed. This method takes any **implementer** of the [EventData interface]({{site.baseurl}}/ApiReference/data/observable/EventData.md) as event data. It includes basic information about an event&mdash;its name as `eventName` and an instance of the event sender as `object`).
```
ErjanGavalji, line 195: Let's make the URL relative (remove the {{site.baseurl}} parameter and replace it with ../../ instead)

core-concepts/events.md
```diff
@@ -302,8 +311,8 @@ var weakEventListenerOptions: weakEventListenerModule.WeakEventListenerOptions =
...
 ```
 
-The previous example shows how to attach a weak event listener to an observable object instance. A closer look to the `handlePropertyChange` function shows that `text` property of the `this` object is changed when the `propertyChange` event is raised (via the button tap event). It demonstrates how to use the `handlerContext` property&mdash;its value is taken as an argument to `this` inside the event handler function.
-
+__Example 7__ shows how to attach a weak event listener to an observable object instance. A closer look to the `handlePropertyChange` function shows that `text` property of the `this` object is changed when the `propertyChange` event is raised (via the button tap event). The function demonstrates how to use the `handlerContext` property&mdash;its value is taken as an argument to `this` inside the event handler function.
+<Comment: What does it refer to at the beginning of the previous sentence? Something specific in Example 7? The function? The text property? The event? It is not clear to me so I do not think the customer will know either. Please replace "it" with the noun that is doing the demonstrating.>
```
ErjanGavalji: Do not forget to remove the comment :)

core-concepts/events.md
erikruth, line 195: I'm not sure if you are asking for the change in the body text or the code snippet or both. I'm not a developer and don't want to introduce an error.

PR #293, opened 2016-03-29
Angular specific changes to action-bar article

vakrilov:
Added angular/nativescript specifics to the ActionBar article


ikoevska:
ping @erikruth

PR #314, opened 2016-04-11
Fixed xml tags typos and removed file extension when working with resources

VladimirAmiorkov:


VladimirAmiorkov:
The broken xml tags lead to unusable code snippets and the inclusion of the extension on the file path when working with resources on an Image tag is actually not supported as mentioned in [this](https://docs.nativescript.org/ui/ui-images#load-images-from-resource) article from the documentation.

PR #328, opened 2016-04-28
Initial commit of the ng2 getting started guide.

tjvantoll:
Also restructuring the guides to live in a new “tutorials” folder.


tjvantoll:
Ping @ikoevska for review if you’re available. Note that this is a new guide and there is a whole lot of content here, so I’ll definitely understand if you’re not able to get to this.

I’m going to land this content as is for now, just because we need these files to be in master to test out these changes, but I’m happy to address any concerns you have in subsequent commits.

PR #332, opened 2016-05-03
TabView for angular article.

nsndeck:
Article for how to use TabView in NativeScript-Angular scenario.

### DIFF
--- a/cookbook/tab-view.md
+++ b/cookbook/tab-view.md
@@ -1,0 +1,128 @@
+---
+title: Tab View
+description: NativeScript for Angular Documentation - Using Tab View
+position: 5
+slug: tabview
+previous_url: /tabview
+environment: angular
+--- 
+
+# Tab View
+
+Using a `TabView` inside Angular 2 app requires some special attention how to provide title, iconSource and content (view) of the TabItem. In a pure NativeScript application `TabView` has an items property which could be set via XML to an array of TabViewItems (basically an array of objects with `title` and `view` properties). However NativeScript-angular does not support nested properties in its html template, so adding `TabViewItem` to TabView is a little bit different. NativeScript-angular provides a custom Angular 2 directive which simplifies the way native TabView should be used. Following is an example of how to add TabView to your page (with some clarifications later):
+
+```XML
+// tab-view-test.html
+<TabView>
+    <StackLayout *tabItem="{title: 'Profile', iconSource: '~/icon.png'}" >
+        <ListView [items]='items'>
+            <template let-item='item'>
+                <Label [text]='item.itemDesc' ></Label>
+            </template>
+        </ListView>
+    </StackLayout>
+    <StackLayout *tabItem="{title: 'Stats'}">
+        <Label text="Second tab item"></Label>
+    </StackLayout>
+    <StackLayout *tabItem="{title: 'Settings'}">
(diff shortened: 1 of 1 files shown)


cookbook/tab-view.md
```diff
@@ -0,0 +1,130 @@
...
+description: NativeScript for Angular Documentation - Using Tab View
+position: 5
+slug: tabview
+previous_url: /tabview
+environment: angular
+--- 
+
+# Tab View
+
+Using a `TabView` control inside Angular 2 app requires some special attention due to the complexity of the NativeScript control like custom item template, bindings and so on. NativeScript-angular plugin provides a custom Angular 2 component which simplifies the way native ListView should be used. Following is an example of how to add TabView to your page (with some clarifications later):
```
hdeshev: The `TabView` component can be customized using templates, and we support that through special directives...

Mentions a ListView? Possibly a copy/paste error. Also "The `nativescript-angular` plugin... component which configures the native control."

cookbook/tab-view.md
```diff
@@ -0,0 +1,130 @@
...
+        for (let i = 0; i < 5; i++) {
+            this.items.push(new DataItem("item " + i));
+        }
+    }
+}
+```
+
+As shown there is nothing complex in a way TabView component is used, but some points need clarifications.
+
+* tabItem - The tabItem tag is used to define a plain json object that defines `title` of the TabViewItem and the element that follows (including the element in which tabItem attribute is used) this template tag will be used as TabItem content.
```
hdeshev: `tabItem` is a directive, and an attribute (not a tag)

cookbook/tab-view.md
```diff
@@ -0,0 +1,130 @@
...
+
+This is a typical usage of the TabView component however if business case requires there are a few options for customizations.
```
hdeshev: This sentence seems superfluous given the next heading. I'd delete it.

cookbook/tab-view.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Customizing Tab View
+
+The most common customization of TabView control is customizing the background color of the selected tab item and using some other than first tab item for start up. Following is how to achive that by little modifications to the previous example.
```
hdeshev: We should mention the way the JSON object works here. As in "customizing `TabItem` instances happens by passing a JSON object containing all the properties that the `tabItem` directive should transfer to the native object. A list of supported properties is available at: ..."

cookbook/tab-view.md
```diff
@@ -0,0 +1,130 @@
...
+        <Label text="Third tab item"></Label>
+    </StackLayout>
+</TabView>
+```
+
+And result is a TabView which selects second tab at start up and uses red color for color of the selected tab.
+
+### Binding TabView selectedIndex
+
+NativeScript-angular TabView `selectedIndex` property can be used in a two-way binding scenarios and using this kind of binding is relative simple. Just use standart `ngModel` syntax to a data model property (for the sake of example TabViewTest class is used as binding context) and set data model property `tabSelectedIndex` to desired value.
```
hdeshev: Typo: standar**d**

hdeshev: Also why name the property `tabSelectedIndex`? It should be `selectedIndex`.

PR #355, opened 2016-05-11
Changed the frame module link to point to the correct location in the…

alexziskind1:
… API Reference
There are a few places like this. Will help change as I come across them.


tjvantoll:
Merged in 92e518f974fb92c8799ae50b95c26aefb3bf1294. Thanks @alexziskind1!

PR #362, opened 2016-05-16
Update running-latest.md

erikruth:
There are three inline comments for you to address.

I also have a general style question. In both the "Behind the curtains of running a NativeScript application" section and the "Building the repo" section, in the "term" list you used hyphens after the first/defined/emphasized term rather than a colon which is used in other NativeScript term lists. I had a discussion with the XAML team and they are moving from hyphens to colons, and prefer to have the first term bold. Is there a reason you chose to use hyphens here?

### DIFF
--- a/releases/running-latest.md
+++ b/releases/running-latest.md
@@ -6,30 +6,30 @@ slug: latest-code
 previous_url: /running-latest
 ---
 
-#Running Latest Code
-<Comment: I think the title and description should say "Running the Latest Code">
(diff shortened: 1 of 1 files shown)


tsonevn:
Thank you for your suggestion about using **colon** rather than a **hyphens**. I have replaced all  hyphens with colon in "Behind the curtains of running a NativeScript application" and "Building the repo" sections. We will keep using **colon** instead **hyphens**

PR #371, opened 2016-05-22
Update Cross-Platform Modules.md

erikruth:
Made many small capitalization changes. I think the "definition" that follows each link should, where it makes sense without causing confusion, begin with a capital letter. I know that in most cases, the definition is the title of the issue that was posted and I tried to respect that, but where the issue title didn't make sense as a definition, I slightly modified it to make more sense.


ErjanGavalji:
[Applied the changes to the origin of the article](https://github.com/NativeScript/NativeScript/pull/2186) (the build takes it from the NativeScript/NativeScript repo).

PR #392, opened 2016-06-07
Update links to point to intended documentation

trevordowdle:
Just noticed that the links could use an update while going through the tutorial.

Thanks


tjvantoll:
Landed in 49cd8bcbfbefa0e558ba36aaa2946408c726d93b. Thanks!

PR #396, opened 2016-06-11
Fix broken Angular doc links

joshgking:
Fixes a bunch of broken links in Angular docs

### DIFF
--- a/core-concepts/modules.md
+++ b/core-concepts/modules.md
@@ -27,72 +27,72 @@ In your project, the files for each module reside in a dedicated subdirectory in
 * 
 ### Core modules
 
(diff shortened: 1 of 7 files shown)


tjvantoll:
Hey @joshgking,

Thanks so much for taking the time to patch these up, but we took a different approach in master with https://github.com/NativeScript/docs/pull/396 that required some additional Jekyll changes. Basically because there are two cookbooks versions, and we want to link to the appropriate one based on whether you’re looking at the {N} core docs or the {N}+ng2 docs.

The master branch will go live with the next {N} release which is coming up soon. Thanks again and sorry I didn’t mention this earlier.

joshgking:
No worries! Thanks for the clarification @tjvantoll

Issue #390, opened 2016-06-07
Cookbook UI documentation + ng2


hdeshev:
_From @tjvantoll on June 6, 2016 16:38_

A lot of the “cookbook” UI documentation on docs.nativescript.org is not applicable to building with ng2. For instance nearly all of the components include a `<Page>` XML element—e.g. http://docs.nativescript.org/angular/cookbook/ui/activity-indicator.

This problem recently came up on Twitter https://twitter.com/weezhen/status/739487960014290944.

I’m not exactly sure how we’d tackle this because the cookbook comes from the modules repo, but I thought I’d least report this so it’s documented.

_Copied from original issue: NativeScript/nativescript-angular#276_

giordy:
I support this! :)

For example I'm having a bit of a hard time in finding out how to use TabViews in ng2.
Thankfully I found a snippet in StackOverflow that gave me some direction, however the cookbook would be golden to get started with components like these.

I should add that I'm still learning ng2 + Typescript + NativeScript and that adds to the confusion.

Thanks for your work! :+1:

tjvantoll:
Hey @giordy,

Make sure to check out our new code samples, which are intended to help solve this: https://docs.nativescript.org/angular/code-samples/overview.html. Feedback is welcome :)

giordy:
Ha!

Had completely overlooked that. Thanks @tjvantoll

jsfgreen:
Any plans to replicate the {N} docs for angular?  Basically include the angular version of each example in the docs.  A daunting task, no doubt, but not because of complexity, more just seems very tedious but very doable.

It would be really nice, because it's not as obvious as you might think to translate the XML template to an angular "HTML" template.

etabakov:
I'm closing the issue. The [code samples] (https://docs.nativescript.org/angular/code-samples/overview) for Angular are the equivalent of the cookbook for NativeScript Core. Regarding the parity of these resources - this is a feedback that we will have in mind with the revamp of the documentation that we are currently working on.

PR #398, opened 2016-06-13
Link to NS cookbook using the new {%ns_cookbook%} helper

hdeshev:
Make sure we use the correct NS cookbook URL's even in the Angular build.


hdeshev:
Merge after telerik/nativescript-docs#25

hdeshev:
Fixes #390

ErjanGavalji:
"Let him who is without sin among you be the first to throw a stone..."

👍

PR #412, opened 2016-06-22
Change status bar style iOS

tsonevn:


core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+There are two options to change the status bar style for iOS applications in NativeScript.
+
+- By changing the `NavigationBar` style. 
+- By using the `Info.plist` file.
+ 
+## Changing Status Bar Style With Help Of NavigationBar barStyle Property
+ 
+This method is easier, but it implies using an `ActionBar`. NativeScript `ActionBar` is common abstraction over iOS `UINavigationBar` and Android's `ActionBar`. It is a bar typically located at the top of the screen and provides title and navigation control in your application.
+
+- Use page's XML file to activate `ActionBar` in your application:
```
hshristov: Declare sound more correct than Activate.

core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+}
+exports.loaded = loaded;
+```
+``` TypeScript
+import { EventData } from "data/observable";
+import { Page } from "ui/page";
+import { topmost } from "ui/frame";
+
+export function loaded(args: EventData){
+    var page:Page = <Page> args.object;
```
hshristov: no need to specify the type because you cast it.
use `let` instead of `var`.

core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+     if (page.ios) {
```
hshristov: don't use `page.ios` - use `platform` module to check the OS.

core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+
+- Changing The Style Of UINavigationBar 
+
+You should use the `frame` module to access the native instance of `UINavigationBar`. Then you can use its `barStyle` property to change its style to `UIBarStyleBlack`.
+
+``` JavaScript
+var frame = require("ui/frame");
+function loaded(args) {
+    var page = args.object;
+    if (page.ios) {
```
hshristov: don't use `page.ios` - use `platform` module to check the OS.

core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+In a result your `Info.plist` should look like that:
+
+``` XML
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+    <dict>
+        <key>CFBundleDevelopmentRegion</key>
```
hshristov: You don't need the whole Info.plist here.
Paste only the relevant section:
```
...
<key>UIStatusBarStyle</key>
<string>UIStatusBarStyleLightContent</string>
<key>UIViewControllerBasedStatusBarAppearance</key>
```

core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+        <string>UIStatusBarStyleLightContent</string>
+        <key>UIViewControllerBasedStatusBarAppearance</key>
+        <false/>
+    </dict>
+    </plist>
+```
+
+![changing-status-bar-style](../img/change-status-bar-style-ios/status-bar-style-info.plist-startup-style-change.png "changing-status-bar-style")
+
+> Now on the place of the status bar we can see one white line. That's happened, because we changed the color of the icon to white, however the background color is the same as those of the icon. To fix this we should set page `backgroundColor` property. You should set also the `backgroundSpanUnderStatusBar` property to `true`. This will span the background color under the status bar:
```
hshristov: Is `>` on purpose? It makes text grey?

core-concepts/change-status-bar-style-ios.md
```diff
@@ -0,0 +1,198 @@
...
+    <Label text="{{ message }}" class="message" textWrap="true"/>
+  </StackLayout>
+</Page>
+```
+ 
+The result is:
+  
+![changing-background-span-under-statusBar-to-true](../img/change-status-bar-style-ios/status-bar-style-info.plist-startup-backgroundSpan.png "changing-background-span-under-statusBar-to-true")
+ 
+> The sample projects for both cases are available in those GitHub repositories: [StyleStatusBariOSviaActionBar](https://github.com/tsonevn/StyleStatusBariOSviaActionBar.git) and [StyleStatusBariOSviaInfo.plist](https://github.com/tsonevn/StyleStatusBariOSviaInfo.plist.git)
```
hshristov: Is `>` on purpose? It makes text grey?

PR #418, opened 2016-07-01
Update for navigation article for angular

vakrilov:
Update angular navigation article with bits for the new angular router.
Depends on NativeScript/nativescript-angular#319


vakrilov:
ping @NickIliev @tsonevn

PR #425, opened 2016-07-04
copy editing with several comments to review

erikruth:
This has several comments. Also, code examples should have SEO-friendly captions so I've added stubs for them but you need to write the caption text.


tsonevn:
Hi @erikruth, 
I have just created new pull request with some edits connected with your comments: [#442](https://github.com/NativeScript/docs/pull/442)

PR #461, opened 2016-07-25
Is a note

jamlfy:
Is create a note

### DIFF
--- a/ui/action-bar.md
+++ b/ui/action-bar.md
@@ -110,7 +110,7 @@ The result is:
 ![title-view-ios](../img/modules/action-bar/title-view-ios.png "title-view-ios")
 ![title-view-android](../img/modules/action-bar/title-view-android.png "title-view-android")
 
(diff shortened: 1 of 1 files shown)


ui/action-bar.md
```diff
@@ -110,7 +110,7 @@ The result is:
 ![title-view-ios](../img/modules/action-bar/title-view-ios.png "title-view-ios")
 ![title-view-android](../img/modules/action-bar/title-view-android.png "title-view-android")
 
-Note that you can use CSS to style the elements inside the `titleView`.
+> Note that you can use CSS to style the elements inside the `titleView`.
```
tjvantoll, line 113: Thanks for pointing this out @alejonext!

I’ll go ahead and land this. But just so you’re aware for future reference, we usually bold the word “Note” within these blocks, so that the rendering looks like this:

> **Note**: You can use CSS to style the element inside the `titleView`.

PR #462, opened 2016-07-26
GridLayout code-behind example added

NickIliev:


ErjanGavalji:
👍 Ping @hshristov for his opinion as well

ui/layout-containers.md
```diff
@@ -276,6 +276,53 @@ Label 3 has a fixed width of 150 pixels. Label 1 is given more space than it act
 
 ![GridLayout](../img/modules/layouts/grid-layout5.png "GridLayout")
 
+
+### Sample (Create grid dynamically via code behind)
+```TypeScript
+import gridModule = require("ui/layouts/grid-layout");
```
hshristov: Lets use `import {GridLayout} from "ui/layouts/grid-layout"`

ui/layout-containers.md
```diff
@@ -276,6 +276,53 @@ Label 3 has a fixed width of 150 pixels. Label 1 is given more space than it act
...
+
+//Grid wrapper
+var grid = new gridModule.GridLayout();
+
+//Create Index Label
```
hshristov: Whats that Index Label? Why not just Label?

ui/layout-containers.md
```diff
@@ -276,6 +276,53 @@ Label 3 has a fixed width of 150 pixels. Label 1 is given more space than it act
...
+let titleLabel = new Label();
+titleLabel.text = "NativeScript";
+grid.addChild(titleLabel);
```
hshristov, line 290: Add note that we add `titleLabel` as child to the grid.
Also note that default `row` and `column` values are `0` so it will go in row 0, column 0.

ui/layout-containers.md
```diff
@@ -276,6 +276,53 @@ Label 3 has a fixed width of 150 pixels. Label 1 is given more space than it act
...
+infoLabel.text = "Truly native mobile apps";
+infoLabel.backgroundColor = new Color("gray");
+grid.addChild(infoLabel);
+
+//Create the share button
+let shareButton = new Button();
+shareButton.text = "Share This!";
+grid.addChild(shareButton);
+
+//Add Grid Rows as if rows="*, 100, auto" cols="250, *"
```
hshristov: Explain the meaning of `*`, `100` and `auto` 
`*` - occupy the remaining available space 
`100` - fixed column width. If elements in this columns wants to be more - we will coerce their width to the column width.
`auto` - the column width will be the width of the widest element in that column.

Same logic for row height.

ui/layout-containers.md
```diff
@@ -276,6 +276,53 @@ Label 3 has a fixed width of 150 pixels. Label 1 is given more space than it act
...
+grid.addRow(new gridModule.ItemSpec(1, gridModule.GridUnitType.star));
+grid.addRow(new gridModule.ItemSpec(100, gridModule.GridUnitType.pixel));
+grid.addRow(new gridModule.ItemSpec(1, gridModule.GridUnitType.auto));
+
+grid.addColumn(new gridModule.ItemSpec(250, gridModule.GridUnitType.pixel));
+grid.addColumn(new gridModule.ItemSpec(1, gridModule.GridUnitType.star));
+
+//Assign views to their row
+gridModule.GridLayout.setRow(titleLabel, 0);
```
hshristov: Explain that we specify in which row/column an element will be. If they are not set - element default row/column is 0/0.

ui/layout-containers.md
tjvantoll: You also need to `import {Color} from "color";`.

tjvantoll: `import {GridLayout, ItemSpec, GridUnitType} from "ui/layouts/grid-layout"`. And you can shorten usage of those three things in the code below too.

ui/layout-containers.md
tjvantoll: Agreed, and as a broader point, all other examples in this file have an image so you can see how the code is rendered. I would recommend adding one for this one too, as currently generic comments like “Index label” and “share button” don’t mean much to the reader.

PR #486, opened 2016-08-30
Fix incorrect ANDROID_HOME installation inctructions

roblav96:
The brew installation creates a symbolic link in `/usr/local/opt/android-sdk` that links to the correct folder the SDK was installed to. `/usr/local/Cellar/android-sdk/24/` is incorrect in my case, because brew installed it in `/usr/local/Cellar/android-sdk/24.4.1_1/` at noted here:
brew install android-sdk

@NathanaelA has mentioned that without the trailing `/` on `export ANDROID_HOME=/usr/local/opt/android-sdk` it will not work because in TNS it reads the env like so:

``` javascript
var someName = $ENV['ANDROID_HOME'] + PathToExe
```

I have had no such issues without this `/` so someone should double check on it.

### DIFF
--- a/start/ns-setup-os-x.md
+++ b/start/ns-setup-os-x.md
@@ -80,17 +80,18 @@ Complete the following steps to setup NativeScript on your OS X development mach
         </code></pre>
 
     1. Install the [Android SDK](http://developer.android.com/sdk/index.html).
-        1. In the terminal, run the following command.
+        1. In the terminal, run the following commands:
 
             <pre class="add-copy-button"><code class="language-terminal">brew install android-sdk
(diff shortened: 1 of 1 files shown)


tjvantoll:
Thanks @roblav96!

This is excellent and simplifies the process so much. I verified that this works on my Mac so I’m going to go ahead and merge this in. Thanks again!

tjvantoll:
Merged in e449a4e9191adf32a4ff45b9db4f98691c423bfe.

roblav96:
@tjvantoll No problem! Glad to help.

PR #487, opened 2016-09-01
weakEventListener renamed weak-event-listener

roblav96:
As per the latest `tns-core-modules`


tjvantoll:
Great catch! Thanks again. Merged with 43e9ab8b038702c378100da3cc8cebbb9b847bec.

PR #489, opened 2016-09-02
Fix installation link

DasAllFolks:
Previously, we were linking to the Angular installation guide (rather than the pure NativeScript installation guide) from the pure NativeScript tutorial.

This fixes the documentation to link to the correct installation guide.


tjvantoll:
Thanks @DasAllFolks!

Merged with 2123030d5876ef920f4707ef9b04ddeb60b801bb.

PR #490, opened 2016-09-03
Fixed CSS @import syntax

roblav96:
Trying

``` css
@import { url('~/your-style.css') }
```

didn't seem to work.

According to this recent article, [using-css-animations-in-nativescript](https://www.nativescript.org/blog/details/using-css-animations-in-nativescript) it says

``` css
@import: "~/css/animate.css";
```

is the correct syntax.

### DIFF
--- a/ui/styling.md
+++ b/ui/styling.md
@@ -413,7 +413,7 @@ If you have doubt about the original font name use [Font Book](https://support.a
 The @import CSS rule allows you to import external CSS from local file, resource or url. These rules must precede all other types of rules.
 
 ```CSS
-@import { url('http://some-domain.com/your-style.css') }
-@import { url('res://your-style.css') }
-@import { url('~/your-style.css') }
+@import: 'http://some-domain.com/your-style.css';
+@import: 'res://your-style.css';
+@import: '~/your-style.css';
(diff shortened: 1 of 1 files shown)


tjvantoll:
Hey @roblav96,

Interesting that the top syntax doesn’t work for you because I’m using it in the getting started tutorial. See `@import { url('~/platform.css') };` on https://docs.nativescript.org/tutorial/chapter-2#24-css.

Could you make sure you’re using `import` at the top of the file and give this one more shot? I haven’t gotten any complaints about this in the tutorial so I’m wondering if this problem isn’t specific to the app you’re working in.

roblav96:
@tjvantoll I just tested it on the demo and you're right, it does work fine. My apologies.

It doesn't work for me when using `nativescript-angular` `RC5`.

tjvantoll:
@roblav96 No worries at all. Odd that this would only be a problem with RC 5, as it was working for me with RC 4. Could you file that as a bug at https://github.com/nativescript/nativescript-angular/issues?

I’m going to close this PR for now as this isn’t really a docs issue. If we broke this syntax we need to fix it 😄 

Thanks again.

PR #492, opened 2016-09-03
Update ng-chapter-5.md

fdnhkj:
Import instead of require for nativescript-social-share plugin. 
See https://github.com/tjvantoll/nativescript-social-share#usage


tjvantoll:
Merged with https://github.com/NativeScript/docs/commit/17e755074ec9886574dffc6ca731242b5af70c6b. Thanks!

PR #493, opened 2016-09-03
Update ng-chapter-4.md

fdnhkj:
Use type inference for `grocery` property in the `ListPage` component (just as for the boolean properties - isLoading and listLoaded).


tjvantoll:
Thanks @fdnhkj! Agreed that this is cleaner.

PR #496, opened 2016-09-07
Update publishing-android-apps.md

dotlens:
For Nativescript apps with angular the information was wrong and the styling was misleading.

### DIFF
--- a/publishing/publishing-android-apps.md
+++ b/publishing/publishing-android-apps.md
@@ -56,7 +56,8 @@ You can publish a NativeScript app in *Google Play* the same way [you would rele
 Both *Package Name*, and *Application Id*, are unique identifiers, provided by you for your app.
  - *Package Name* is used to identify resources such as the `R`.
  - *Application Id* is used to identify your app on devices and at the *Google Play*.
-In the NativeScript framework, both are set to the `nativescript.id` in `package.json`.
+
+In the NativeScript framework, both are set to the `applicationId` in `app.gradle`.
 The NativeScript CLI build system will set them as the `package` attribute in the generated project in `platforms/android/src/main/AndroidManifest.xml`.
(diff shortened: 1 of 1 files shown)


tjvantoll:
Hey @dotlens,

Sorry for the late reply here. Could you provide a little more information on the problem that you ran into? In theory a NativeScript app’s `nativescript.id` in `package.json` and `android.defaultConfig.applicationId` in `app.gradle` should match up, and changing the value in the `package.json` is the recommended approach.

I just gave that a shot in a new Angular app and it worked for me as expected.

dotlens:
Hey @tjvantoll 

Indeed it would make sense to change it in the package.json file. But I just tested it with two nativescript-angular projects of mine and it did not work. When I changed the package name in "app\package.json" there was no effect. I had to change it in "app\App_Resources\Android\app.gradle".

I did it as it was proposed:
1. Change nativescript.id in the package.json
2. remove the "platforms\android" folder
3. run "tns prepare android"
4. run "tns run android"
5. check "platforms\android\build\intermediates\manifests\full\F0F1\debug\AndroidManifest.xml" --> not correct
6 check deployed package name --> not correct

Then I tried it with the app.gradle file:
1. Change the "android.defaultConfig.applicationId" in the "app\App_Resources\Android\app.gradle"
2-6 as before --> correct output

I hope this information could help you

Btw: The formatting of the text was also not good. There was no separation between the second bullet point and the following text which is confusing.

Cheers Claudio

PS: Im using tns-core-modules 2.3.0 and nativescript-angular 1.0.1

tjvantoll:
Thanks for the detail @dotlens—much appreciated, and I was able to recreate.

I actually think this is a bug in our CLI so I opened up a discussion there: https://github.com/NativeScript/nativescript-cli/issues/2089. Regardless, either our CLI needs to be updated or our docs. Thanks for reporting this; we have to figure out which is correct now 😄

NickIliev (approved):

PR #497, opened 2016-09-10
Add platform qualifier bundling instructions

roblav96:
If there's a better way to do this, let me know.

Some platform qualifier modules were bundled correctly without this, but most were not when using `ng2-rc6`.

> im importing `import {Permissions} from "../../shared/permissions/permissions"` which is an injectable service, but i have `permissions.ios` and `permissions.android`. it looks like its bundling the `permissions` file, which is dummy typescript declarations, and ignoring the platform file patterns
> 
> `[667] ./shared/permissions/permissions.js 287 bytes {0} [built]` but it SHOULD be bundling `./shared/permissions/permissions.ios.js`
> 
> @walkerrunpdx yes but i have `import "./shared/delegates/delegate"` in my `main.js` which is really `/delegate.ios` and in the bundler it reports `[682] ./shared/delegates/delegate.ios.js 1.38 kB {0} [built]` so thats working
> 
> plus the file size of `permissions.ios.js` is certainly not 287 bytes. thats the size of the dummy declarations file

### DIFF
--- a/tooling/bundling-with-webpack.md
+++ b/tooling/bundling-with-webpack.md
@@ -126,6 +126,38 @@ global.loadModule(myPlugin);
 
 Webpack bundling can fail for different reasons, or it can generate code that breaks at runtime (like the dynamic imports scenario outlined above). Debugging problems is easy once you know what code runs on your device.
 
+### Enabling Platform File Extensions
+
+Say you have the following files:
+
+- `permissions.ios.ts`
+- `permissions.android.ts`
+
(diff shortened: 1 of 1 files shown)


hdeshev:
Thanks for your contribution! I realize that documenting this feels like a suboptimal approach (as opposed to solving the issue in nativescript-dev-webpack), but for now it's better than nothing.

I am working on a change to module resolution for `nativescript-dev-webpack` that already takes care of `.android.js` and `.ios.js` extensions, but it may require some changes to user projects and is not ready to be released yet.

PR #259, opened 2016-03-08
Document using sandbox-pod

teobugslayer:

PR #498, opened 2016-09-11
Restoring cocoapods docs on sandbox-pod removed by ns-bot

dennistang:
Just had issues with `tns doctor` where sandbox-pod was having permission issues when testing pod install.

Saw these docs were added at some point but as mentioned it looks like ns-bot had removed it for whatever reason: https://github.com/NativeScript/docs/pull/259

These docs would've saved me a couple hours of debugging, so hopefully adding these docs back in will stick this time around?


dennistang:
Closing, going to submit a PR to perhaps default `USE_POD_SANDBOX` to true.

PR #503, opened 2016-09-16
zh-CN (i18n & L10n)

xgqfrms-GitHub:
# zh-CN (i18n & L10n)


tjvantoll:
Hey @xgqfrms-GitHub,

Thanks so much for starting this up. Unfortunately though, full documentation translations isn’t something that we can maintain at this time in this repo. You can submit that idea as something we should support on https://nativescript.ideas.aha.io/.

I would recommend getting this started as a fork of the main docs repo. We can then link to your translated versions until we have the capacity to handle this in the main project.

PR #506, opened 2016-09-17
Fix incorrect class name

amjd:
The example for italic fonts had the class name `font-weight-bold` instead of `font-italic`.


tjvantoll:
Closed with c64ba3bc60fc3421ea05e4609321ab9d366c921f. Thanks @amjd!

PR #521, opened 2016-09-29
added Pseudo-selector highlighted in Styling >> Selectors

NickIliev:


ErjanGavalji (commented):
Do we need nativescript and angular-specific snippets if both are the same?

PR #524, opened 2016-09-30
remove quotation marks from the CSS property table in background-image property description

tsonevn:


tjvantoll:
Hey @tsonevn,

Is there a reason these quotes need to be removed? Most of the other values in the same section also use quotes. See http://docs.nativescript.org/ui/styling#supported-css-properties.

I actually think the appropriate thing to do would really be placing all of these things in `<code>` tags. So `repeat`, `repeat-x`, etc. Thoughts?

tsonevn:
Hi @tjvantoll,
It will be more clear for the people, who read the documentation. In case you  set `background-repeat` and selected option is with quotes in the CSS. The CSS attribute will be not attached correctly. this is the reason I removed them. However if you think that this is not good idea, I will close this Pull request.

tjvantoll:
@tsonevn What are your thoughts on using this as an alternative approach: https://github.com/NativeScript/docs/pull/534. You can see what this looks like rendered at https://github.com/tjvantoll/docs/blob/c2b98bf9a9f23301b23ecb17a6f64791cbff9073/ui/styling.md#supported-css-properties.

tsonevn:
Hi @tjvantoll,
That is great idea. I will close my pull request and you could merge yours. Regarding to that, could you merge your request to `master` branch instead of `production`?

tjvantoll:
@tsonevn Yep, sounds good. I’ll merge it into `master` right now.

PR #526, opened 2016-10-01
Fixed ngModelChange output syntax

yjaaidi:
One should use (ngModelChange) to detect the model change (this only works if [ngModel] is set).
Remember that `[(x)]="value"` is a two-way binding shortcut for `[x]=value (xChange)="value = $event"` .


tjvantoll:
Merged in with 681f1a1f9c714a02afd8b6195c64b5d7e225e9af. Thanks @yjaaidi!

yjaaidi:
You're welcome!

PR #562, opened 2016-11-13
Updated ns-setup-os-x.md for Firebase-Usage

Taremeh:
If you want to use Firebase with AVD, you´ll need to specify ANDROID_HOME as "/usr/local/Cellar/android-sdk", otherwise it won't recognise Google APIs Intel Atom as the CPU/ABI, which is needed to run Google Play Services (which again is needed for Firebase).

More on this issue:
https://github.com/EddyVerbruggen/nativescript-plugin-firebase/issues/207

### DIFF
--- a/start/ns-setup-os-x.md
+++ b/start/ns-setup-os-x.md
@@ -87,7 +87,7 @@ Complete the following steps to setup NativeScript on your OS X development mach
 
         1. Next, run the following command to set the ANDROID_HOME system environment variable:
 
-            <pre class="add-copy-button"><code class="language-terminal">export ANDROID_HOME=/usr/local/opt/android-sdk
+            <pre class="add-copy-button"><code class="language-terminal">export ANDROID_HOME=/usr/local/Cellar/android-sdk
             </code></pre>
 
             <blockquote><b>NOTE</b>: This is the directory that contains the <code>tools</code> and <code>platform-tools</code> directories.</blockquote>


tjvantoll:
Hey @Taremeh,

Sorry for the late response here. This is very interesting. I’m a little hesitant to include this, as most information I’ve seen on the web seems to recommend sticking with `/usr/local/opt` (for example http://stackoverflow.com/questions/35337601/why-is-there-a-usr-local-opt-directory-created-by-homebrew-and-should-i-use-it).

Personally I’d rather see this as a note that’s included in the Firebase plugin’s README, but I’m definitely not a Homebrew expert so I’ll leave this open for others to chime in.

dtopuzov:
/usr/local/opt/android-sdk is actually symlink to folder where Android SDK is installed.
```
/usr/local/opt/android-sdk -> ../Cellar/android-sdk/24.4.1_1
```

At the moment is is /usr/local/Cellar/android-sdk/24.4.1_1, but we have no guaranty it will be same in future.
More over, just setting ANDROID_HOME to /usr/local/Cellar/android-sdk will fail, because actual location is /usr/local/Cellar/android-sdk/24.4.1_1 

I believe problem with Google API CPU/ABI, is not related to ANDROID_HOME.

Here is how you can create accelerated Google API emulator:
```
echo y | $ANDROID_HOME/tools/android update sdk --filter android-23 --all --no-ui
echo y | $ANDROID_HOME/tools/android update sdk --filter sys-img-x86-google_apis-23 --all --no-ui
echo no | $ANDROID_HOME/tools/android create avd -n Emulator-Api23-Google -t android-23 --abi google_apis/x86 -c 12M -f
```

tjvantoll:
Closing based on @dtopuzov’s comments.

PR #579, opened 2016-11-18
Update quick-setup.md

msangapu:
Added note on how to handle admin privileges on windows


dtopuzov:
If you install something as Admin, most likely you should always use it as Admin later.
Not sure if is convenient, may be we can add this update, but we should also state that installation with Admin will cause inconvenience in usage later.

msangapu:
I submitted this request as it would not let me proceed without admin privileges.

PR #585, opened 2016-11-21
Remove a sentence that might confuse new users

tjvantoll:
Ping @donangel.

I’d like to remove this sentence because it has some grammatical issues, and because I think the phrase “all mobile SDKs” may confuse new users that don’t know _which_ mobile SDKs to install.

Could you provide a little more detail on what confused you about this documentation before? We can probably rephrase this or add something different to address your concerns. I also asked the CLI team whether the setup scripts could install Node+npm+tns, which would simplify these instructions considerably.

Thanks.


donangel:
Thanks for the follow up.

Today, I cannot remember where was the confusing part, when I modified the documentation. It was a while, when I created this request and I'm a bit embarrassed now for not remembering. 

I've a hunch that when installed {N} on a brand new system without any SDK, there was something, which pushed me to modify these docs. But the thing is that until I retry this (and I cannot do it now), I can't tell where the problem was.

Maybe, as you propose, it's worth adding more details to this part of the documentation, instead of removing anything. 

I apologize for not detailing enough my request at that time, I should have done better. Please feel free to either decline it, or to do differently.

Best regards,
Doncho

tjvantoll:
Hey @donangel,

No worries at all. I’m going to go ahead and remove this sentence for now, but I [opened up a conversation on how we can improve this flow](https://github.com/NativeScript/nativescript-cli/issues/2252). Long term I think our installers will (hopefully) make this a non issue. Time will tell 😄

tjvantoll:
Closed with 722445dbf4ecb8db87aa8c51828c0a4d2437a408.

PR #591, opened 2016-11-23
Updated quick-setup with Windows-specific instructions for intel haxm.

ghost:
I followed the installation for Windows on Nativescript today and had to do a little digging to figure out why it wasn't working.

Turns out, when Android SDK says Intel HAXM is installed, it really means it's been downloaded to the extras folder.  Unless the installer is manually run on Windows, Intel HAXM won't be installed, and the default AVD will not run.  I've had to do this on my Windows 10 machine.

%ANDROID_HOME% should be the same environmental variable that the Android SDK has defined as the SDK Path.

Reference: https://software.intel.com/en-us/android/articles/installation-instructions-for-intel-hardware-accelerated-execution-manager-windows

### DIFF
--- a/start/quick-setup.md
+++ b/start/quick-setup.md
@@ -64,6 +64,11 @@ If you’re on Windows, copy and paste the script below into your command prompt
 
 During installation you may need to accept a User Account Control prompt to grant the script administrative privileges. Also, be aware that the script downloads and installs some big dependencies—so it’s common for the script to take a while to complete. When the script finishes, close and reopen your command prompt.
 
+While Intel HAXM is downloaded by the above installation script, Windows users must manually run the installer. To do this, run the following command in your command prompt:
+
+<pre class="add-copy-button"><code class="language-terminal">%ANDROID_HOME%\extras\intel\Hardware_Accelerated_Execution_Manager\intelhaxm-android.exe</code></pre>
+
+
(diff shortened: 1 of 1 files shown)


dtopuzov:
I see we try to install HAXM silently:
https://github.com/NativeScript/nativescript-cli/blob/master/setup/native-script.ps1#L113

Script will first ask you "Do you want to install Android emulator?".
If yes -> "Do you want to install HAXM (Hardware accelerated Android emulator)?"
If yes -> Download the installer and run silent setup.

May be you answered no to "Do you want to install Android emulator?" and in this way you where never prompted to install HAXM or silent installer failed for some reason.

Not sure if we should fix docs or setup script in this case.

ghost:
The only thing I answered "No" to was the install of Google Chrome, which I already had.  I was prompted to install HAXM and answered yes. It showed up as "Installed" in my SDK Manager when I was troubleshooting, but when I ran the AVD it said there was an error and to check that HAXM was properly installed.  I'll try running the install on another computer and see if I encounter the same problem and if silent_install.bat is the issue.

ghost:
The problem is with silent_install.bat.  Both during the nativescript setup script and when I try to manually run silent_install.bat, I receive the printed error "intelhaxm-android.exe is missing", despite it being present in the same folder.  I'm currently checking if replacing silent_install.bat in the script with intelhaxm-android.exe directly will cause any issues. (I don't know why the image is upside down, but it's rightside up when you click it.)

ghost:
Replacing silent_install.bat with intelhaxm-android.exe works. It opens up the installer and the script progresses appropriately afterwards.  It could be a sufficient workaround to silent_install.bat not working.  If that route works better, I can submit another PR with that edit.

Interestingly, after this workaround finishes a successful installation, uninstalling HAXM via `silent_install.bat -u` and then running `silent_install.bat` afterwards successfully reinstalls it. `silent_install.bat -v` confirms both uninstallation and reinstallation.

NickIliev (approved):

PR #598, opened 2016-11-26
typo: Android should be iOS

sebastianovide:


valentinstoychev:
Thank you @sebastianovide!

PR #628, opened 2016-12-15
updated supported Node versions

NickIliev:
As Angular-2 projects are not working with 4.x and currently there are some issues related to usage of 7.x

### DIFF
--- a/start/ns-setup-linux.md
+++ b/start/ns-setup-linux.md
@@ -19,7 +19,7 @@ This page contains a list of all system requirements needed to build and run Nat
 ## System Requirements
 
 * Ubuntu 14.04 LTS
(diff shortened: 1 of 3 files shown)


tsonevn (approved):

realtebo:
Can you tell us, please, of which issues are you telling? 

We're working with v7 and are interested if any trouble are coming to happening

NickIliev:
@realtebo on some Windows PCs with Node.js 7.x the npm is refusing to install nativescript or/and some other npm packets which are dependencies. We are investigating the issue

realtebo:
@NickIliev : i'm aware of problem only with node-gyp, that is simply resolved pre-installing the npm package windows-build-tools. But Thanks for the reply, I'll try to downgrade our apps to node 6.9.x

PR #632, opened 2016-12-26
reduce arrays response using rxjs functional programing

mrsoto:
* remove side effects in foreach using pure functions
* apply switch map


NickIliev (commented):
Great PR! Can you do it against the master branch and not against production as this is the workflow for committing new changes in the docs section

PR #633, opened 2016-12-27
Update ng-chapter-2.md

betosalvador:
Referencing correct file


NickIliev (commented):
In the example shown the HTML part is within the *.ts file as we are using template and not templateUrl

betosalvador:
sorry.  my fault. :(

PR #640, opened 2016-12-31
reduce arrays response using rxjs functional programing

mrsoto:
* remove side effects in foreach
* apply switch map


tjvantoll:
Hey @mrsoto,

Sorry for the super late reply here, and thanks for contributing this change. I’m really hesitant to include this because it’s going to be really hard for the average reader (including me) to parse and read.

I’d prefer to keep this example simple even if it’s not the ideal way of accomplishing the task.

mrsoto:
You are welcome.
   Hope you solve how the original code incorrectly handle multiples requests when response arrive in incorrect order.

PR #643, opened 2017-01-04
Update chapter-2.md

crezione1:


tjvantoll:
Hey @crezione1,

This one also seems to be correct. It looks like we’re using a `views` folder (see https://github.com/NativeScript/sample-Groceries/tree/start/app). If I’m missing something let me know.

PR #644, opened 2017-01-04
Update chapter-2.md

crezione1:


tjvantoll:
Hey @crezione1,

Thanks, but the file is called `app.js` for {N} JavaScript applications, so this should be correct. If I’m missing something let me know.

PR #665, opened 2017-02-01
Homebrew no longer supports the LTS named packages

toddanglin:
Updated tip for macOS to instead suggest that Homebrew users run brew update and brew upgrade node to get the latest version.


tsonevn:
Hi @toddanglin 
I tested `brew upgrade node` command on my side, however, It seems that the node will not be updated successfully. This command will return  `Error: node not installed`.

toddanglin:
I think this command will only work if you originally installed node via Homebrew.

To that end, if the tip needs to address scenarios where node is not installed, changing `brew upgrade` to `brew install` should work. Most important point is that current docs point at a package that no longer exists (`homebrew/versions/node6-lts`).

tsonevn:
Hi @toddanglin,
We encouraged using LTS Node.js version. At the moment we are not sure that we are fully compatible with node 7. Regarding that, we made the needed changes in the documentation with PR: [#680](https://github.com/NativeScript/docs/pull/680)

PR #686, opened 2017-02-24
Update ng-chapter-3.md

jacostaperu:
on Line 709 Where it says:
To use ... app/login/login.component.ts, ...
should be:
To use ... app/pages/login/login.component.ts, ...


tjvantoll:
Merged with 223c91a87a14aa8de8ec006e738accb2309bbcdf. Thanks @jacostaperu!

PR #689, opened 2017-02-26
How to add event in existing element in html?

dinesh36:
Example 1 (JavaScript): Adding an event handler or an event listener using the short and full syntax.

In this section you have explained how to add event to the button. But how to add this button to my view. Or there is any way I can add event to the existing element in my view[html file]?


tjvantoll:
Hey @dinesh36,

Thanks for the feedback. Can you please open this as an issue on https://github.com/NativeScript/docs/issues instead of this pull request?

Thanks!

PR #705, opened 2017-03-17
refactor(setTextFieldColors): Change let to const

Nikhil22:


tjvantoll:
Hey @Nikhil22,

Thanks for this, and sorry for the late reply. I’m going to hold off here, only because we’re not using `const` anywhere in this app right now, so I think it’d be a bit weird for it to be used in only one function.

PR #741, opened 2017-04-25
fixed syntax error in the snippets

NickIliev:


NickIliev:
@toddanglin I've revised this section adding ~/ instead of just ~ . Can you verify that this was the original idea behind [this PR](https://github.com/NativeScript/docs/pull/723/files)

toddanglin:
No...there are two syntaxes supported now:

- ~ (no slash): Will search for SASS imports relative to node_modules folder
- ~/ (with slash): Will search for SASS imports relative to app root

The "no slash" syntax is meant to be consistent with other webpack/Angular import syntaxes. Does that make sense?

sis0k0:
@toddanglin, the ~ (no slash) import doesn't work for me (newly created app with NS 3.0)

toddanglin:
@sis0k0 Do you have the latest nativescript-dev-sass plugin installed?

sis0k0:
Oh, I got it. So, the first section of the article (plain css) should be with `~/` and the sections about using the theme with sass plugin should be just with `~`. Do you agree?

toddanglin:
Ah! I see...yes...my PR may not be valid for the "plain CSS" sections of this article. We specifically need to update the SASS related snippets.

For the plain CSS, I'm not certain "~/" works, either. Don't think that is a built-in short-cut. Perhaps the original, unmodified syntax works fine for plain CSS? Need to re-test locally. Good catch.

Do you want me to submit a different PR for the SASS-specific parts of the article?

NickIliev:
Hey @toddanglin yes please create a PR with the recommended syntax for SASS (for plain CSS "~/" works but perhaps we should revert to the original syntax)

NickIliev:
@toddanglin, @sis0k0 I've just revised this example preserving the CSS syntax (tested on my side and works as expected with ~/) and at the same time revising all the SASS example to use only ~. Can you please verify this PR

sis0k0 (approved):

toddanglin:
Sounds good. Thanks!

PR #758, opened 2017-05-04
Fix for Sign Up Button on Chapter 3.4 Services

kurapikats:
When the user press the Sign Up Button it calls login() but it doesn't do anything.
To fix this: set the this.isLoggingIn property to false so that signUp() can be called.

### DIFF
--- a/tutorial/ng-chapter-3.md
+++ b/tutorial/ng-chapter-3.md
@@ -371,6 +371,7 @@ submit() {
 }
 login() {
   // TODO: Define
+  this.loggingIn = false;
 }
 signUp() {
   this.userService.register(this.user);


tjvantoll:
Hey @kurapikats,

Thanks for the contribution, but this change shouldn’t be necessary, as in section 3.4 the instructions say to click the bottom “Sign up” button, and then the top “Sign up” button which calls register. Now that I’m typing this out though I’m realizing how confusing this probably is. I think a better solution is probably renaming at least one of these buttons to avoid the ambiguity.

tjvantoll:
I’m going to close this one for now, but I added button renaming to the list of revisions that need to happen in the next version of this tutorial.

Thanks again @kurapikats

PR #759, opened 2017-05-05
Convert submodules to subdirectories

etabakov:


PetyaSotirova (approved):
The two README files seem redundant. The rest looks OK and I'd like to take another look before we deploy.

PR #762, opened 2017-05-08
docs(tutorial): use relative urls with moduleId instead of absolute

sis0k0:


tjvantoll:
Thanks @sis0k0! The code changes look fine here. The only thing I think that needs to be added is a short explanation of the `moduleId` property the very first time it’s used. There’s a section right below the first component that talks about the `selector` property, and ideally the `moduleId` explanation would go there.

I can add that if you’d like to save you some time. It’s getting late in the day for me right now so I’ll check back in on this one tomorrow.

tjvantoll:
Hey @sis0k0,

Sorry for the late reply here. As I was researching how to describe the `moduleId` property I came across this in the [Angular changelog](https://angular.io/docs/ts/latest/guide/change-log.html#!#sts=All%20mention%20of%20moduleId%20removed.).

> “You no longer need to write @Component({ moduleId: module.id }), nor should you.”

Given that should we change our recommendation?

sis0k0:
You no longer need to write module ids for the components in Angular in the web, because it's handled by a systemjs loader. Since we don't use systemjs, we still need module id's to resolve relative template and style urls in the components.

sis0k0:
I added a simple explanation about the `moduleId`. Do you think we can add a few more words why we is it not required on the web?

tutorial/ng-chapter-3.md
```diff
@@ -343,9 +344,10 @@ After that, add a new `providers` property to the existing `@Component` decorato
 ``` TypeScript
 @Component({
   selector: "my-app",
+  moduleId: module.id,
   providers: [UserService],
-  templateUrl: "pages/login/login.html",
-  styleUrls: ["pages/login/login-common.css", "pages/login/login.css"],
+  templateUrl: "./login/login.html",
+  styleUrls: ["./login/login-common.css", "./login/login.css"],
```
tjvantoll, line 350: The `/login` part of these paths can go away, correct? (This applies to a few different lines in this PR.)

tjvantoll:
> In NativeScript we use relative paths for the components' assets (such as stylesheets). That's why we also need to set a `moduleId` for our component. We set it to `module.id` - a value provided by the [CommonJS module](http://wiki.commonjs.org/wiki/Modules/1.1).

Is it fair to say that the `moduleId` property needs to point at the id of a JavaScript module that subsequent paths should be resolved against?

> I added a simple explanation about the moduleId. Do you think we can add a few more words why we is it not required on the web?

I would but I don’t think I know the answer 😄 

To ask a related question: is there a compelling reason why we want or need people to use `moduleId` in their components?

Sorry for all the questions here, but I do think this has the potential to confuse readers, especially if they google “moduleId” and find a big bold thing on Angular’s site telling them not to use it.

NickIliev:
Resolved via [https://github.com/NativeScript/sample-Groceries/pull/275](https://github.com/NativeScript/sample-Groceries/pull/275) and [https://github.com/NativeScript/docs/pull/942](https://github.com/NativeScript/docs/pull/942)

PR #770, opened 2017-05-10
Fixed incorrect instructions in Chapter 3

AsixJin:
var user = new Observable({
    email: "user@domain.com",
    password: "password"
});

The above code doesn't work. The code below will get the results desired from this chapter.
var user = new Observable();
user.email = "user@domain.com";
user.password = "password";

### DIFF
--- a/tutorial/chapter-3.md
+++ b/tutorial/chapter-3.md
@@ -206,10 +206,9 @@ Add the following code to the top of `app/views/login/login.js`. The code gets a
 ``` JavaScript
 var Observable = require("data/observable").Observable;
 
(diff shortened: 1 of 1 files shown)


tjvantoll:
Thanks so much for reporting this @AsixJin! We’re pushing people toward using `observableModule.fromObject()`, so I just pushed a commit to address that (https://github.com/NativeScript/docs/commit/f57dca9f85365a3a30c5ab0d835652c64bbdd3cf).

Thanks again!

AsixJin:
No problem! The guide is very helpful!

PR #776, opened 2017-05-15
Article about Image performance techniques in Android

NickIliev:
Related to [https://github.com/NativeScript/NativeScript/issues/2571#issuecomment-299120495](https://github.com/NativeScript/NativeScript/issues/2571#issuecomment-299120495)

### DIFF
--- a/ui/images-optimisations.md
+++ b/ui/images-optimisations.md
@@ -9,20 +9,19 @@ slug: images-performance
 
 One of the most common scenarios for modern mobile applications is to work with multiple images often in high definition formats.
 It is crucial for each mobile developer to handle memory related issues and optimize the application so it could handle large data (e.g. API call which will load hundreds of photos). 
-In this article, we will take a look at how Image module works in NativeScript and 
-cover the techniques that will improve our Android application performance.
 
-# Images optimization techniques for Android
+In this article, we will take a look at how Image module works in NativeScript and 
+cover the techniques that will improve Android application performance.
 
 ## Handling large images and avoiding Out Of Memory exception
 
-Sometimes when working with multiple large images on devices with low memory, we can get Out Of Memory exception. To prevent that scenario, in NativeScript 2.5.x and above using the `src` property in Android will internally load the Bitmap in Java. Bitmap memory stays in Java world and reclaims once the Bitmap is no longer in use (e.g. there is no need for the Javascript object to be collected).This way Bitmap memory management is not an issue.
+In some cases when working with multiple large images on devices with low memory, an `Out Of Memory` exception can occur. To prevent that scenario, in NativeScript 2.5.x and above using the `src` property in Android will internally load the Bitmap in Java. Bitmap memory stays in Java world and reclaims once the Bitmap is no longer in use (e.g. there is no need for the Javascript object to be collected).This way Bitmap memory management is not an issue.
 
 In contrast, when using `ImageSource` or Base64 encoded string, the Bitmap is transferred to Javascript, so it will be released when Javascript object reclaims. Javascript garbage collection happens less frequently than Java garbage collection which might lead to Out Of Memory.
 
 > **Tip**: Use `src` property of your `Image` to set your images to avoid Out Of Memory related issues.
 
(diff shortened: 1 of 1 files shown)


ui/images-optimisations.md
```diff
@@ -0,0 +1,42 @@
...
+One of the most common scenarios for modern mobile applications is to work with multiple images often in high definition formats.
+It is crucial for each mobile developer to handle memory related issues and optimise the applicaiton so it could handle large data (e.g. API call which will load hundreds of photos). 
+In this article, we will take a look at how Image module works in NativeScript and 
+cover the techniques that will improve our applicaiton performance.
+
+# Images optimization techniques for Android
+
+## Handling large images and avoiding OOM (Out of Memory) exception
+
+Sometimes when working with multiple large images on devices with low memory we can get OOM exception. To prevent that scenario, in NativeScript 2.5.x and above using the `src` property in Android will internally load the bitmap in Java. This way Bitmap memory management is not an issue.
```
hshristov: Bitmap memory stays in Java world and is reclaimed once the Bitmap is no longer in use (e.g. there is no need for Javascript object to be collected).

ui/images-optimisations.md
```diff
@@ -0,0 +1,42 @@
...
+
+In contrast, if we use `ImageSource` instance and set it to `imageSource` property of `Image` 
+or if we pass Base64 encoded string, we will load the bitmap in JavaScript which may cause OOM exception due to the synchronization of the garbage collectors (JavaScript and Java).
```
hshristov: When using ImageSource or Base64 encoded string the `Bitmap` is transferred to Javascript so it will be release only when Javascript object is reclaimed. Javascript garbage collection happens less frequently than Java garbage collection which leads to OOM.

ui/images-optimisations.md
```diff
@@ -0,0 +1,42 @@
...
+## Handling large images and avoiding OOM (Out of Memory) exception
+
+Sometimes when working with multiple large images on devices with low memory we can get OOM exception. To prevent that scenario, in NativeScript 2.5.x and above using the `src` property in Android will internally load the bitmap in Java. This way Bitmap memory management is not an issue.
+
+In contrast, if we use `ImageSource` instance and set it to `imageSource` property of `Image` 
+or if we pass Base64 encoded string, we will load the bitmap in JavaScript which may cause OOM exception due to the synchronization of the garbage collectors (JavaScript and Java). 
+
+> **Recommendation tip**: Use `src` property of your `Image` to set your images to avoid OOM related issues.
+
+As an additional feature for Android, NativeScript supports `decodeWidth`, `decodeHeight` and `loadMode` properties. The first two will downsample your image so it will take less memory. With [loadMode](http://docs.nativescript.org/api-reference/modules/_ui_image_.html#loadmode) set to `async` the image will load asynchronously meaning UI won't be blocked by the decoding and preloading operations.
```
hshristov: `loadMode` is supported in both android & ios. `decode*` properties are android only for the moment.

hshristov: All `http` requests are loaded `async` no matter the value of `loadMode`.

ui/images-optimisations.md
```diff
@@ -0,0 +1,42 @@
...
+```XML
+<StackLayout>
+    <Image src="{{ someExtremelyLargeImage }}" decodeWidth="400" decodeHeight="400" loadMode="async" />
+    <Label text="With loadMode set to async the UI won't be blocked" textWrap="true" />
+</StackLayout>
+```
+
+> **Important**: When the `src` value starts with `http` it will be loaded asynchronously no matter what value we have set to `loadMode`.
+
+`Image` will use internal memory and disk cache, so images that are loaded are stored in the memory cache and if they are not needed anymore, they are saved in the disk cache. This way the next time we need the same image NativeScript will load it from memory or the disk cache.
```
hshristov: internal memory and disk cache - implemented only for android at the moment.

hshristov (changes requested):

ui/images-optimisations.md
NickIliev: the whole article section is explicitly for Android - added in the final Note disclaimer and in the title itself

ui/images-optimisations.md
NickIliev: disclaimer added few lines below

PetyaSotirova (approved):
Can we add links to the image API and/or the cookbook? Maybe in a See Also section at the end.
Looks great.

hshristov (approved):

PR #792, opened 2017-05-30
Fix components references from the Angular Components page

tsonevn:
Fix components references from the Angular Components page


PetyaSotirova:
The Components page is shared between the Core and Angular docs and this change will break the links in Core. I think we may need two separate environment-specific articles to solve this one.

PetyaSotirova (approved):

PR #801, opened 2017-06-08
Property system in NativeScript 3.x.x.

NickIliev:
Article about the property system and creating/migrating UI plugin in NativeScript 3.0.0.
Source code based on [this demo plugin](https://github.com/NickIliev/nativescript-my-button).

### DIFF
--- a/core-concepts/properties.md
+++ b/core-concepts/properties.md
@@ -10,515 +10,379 @@ environment: nativescript
 
 # Properties
 
-
 This article contents:
 
 * [Overview](#overview)
-* [Dependency properties](#dependency-properties)
-    * [Introduction](#intro1)
-    * [Declaring a dependency property](#declaring)
-    * [Adding a changed callback](#changedCallback)
-    * [Adding a validation callback](#validationCallback)
-    * [Creating an inheritable dependency property](#inheritable)
-* [Style properties](#style-properties)
-    * [Introduction](#intro2)
-    * [Setting a style property](#setting)
+* [Prerequisites](#prerequisites)
+* [Basic UI plugin structure](#basic-ui-plugin-structure)
+* [Property class](#property-class) 
+* [CssProperty class](#cssproperty-class) 
+* [Registering the Property](#registering-the-property) 
+* [Platform-specific code](#platform-specific-code) 
 
 ## Overview
 
-Being a TypeScript framework, NativeScript uses TypeScript properties. After transpilation, these result in ECMAScript v.5 compliant JavaScript with setter and getter methods to support working with class members, thus ensuring readable and manageable code. The code in __Example 1__ demonstrates how the TypeScript is transformed to JavaScript.
+Being a TypeScript framework, NativeScript uses TypeScript properties. After transpilation, these result in ECMAScript v.5 compliant JavaScript with setter and getter methods to support working with class members, thus ensuring readable and manageable code. This article will guide you through the process of creating basic UI plugin using the property system. All the code is in TypeScript and requires NativeScript 3.x.x or newer version.
 
-> **Important:** The techniques described in this article are applicable for NativeScript versions 2.5.x and below.  
+## Prerequisites
 
-### Example 1: How TypeScript is transformed to JavaScript.
-``` JavaScript
-var MyClass = (function () {
-  function MyClass() {
-  }
-  Object.defineProperty(MyClass.prototype, "myProperty", {
-    get: function () {
-      return this._myProperty;
-    },
-    set: function (value) {
-      this._myProperty = value;
-    },
-    enumerable: true,
-    configurable: true
-  });
-  return MyClass;
-})();
-exports.MyClass = MyClass;
+- NativeScript 3.x.x or newer version
+- Plugin structure as follows
+
 ```
-``` TypeScript
-export class MyClass {
-  private _myProperty: number;
-  public get myProperty(): number {
-    return this._myProperty;
-  }
+src/
+├── index.d.ts
+├── my-button.d.ts
+├── my-button-base.ts
+├── my-button.ios.ts
+├── my-button.android.ts
+├── package.json
+├── references.d.ts
+├── tsconfig.json
+└── platforms/
+    ├── android/
+    │   ├── include.gradle
+    └── ios/
+        └── Info.plist
+```
+> Note: For more details about the plugin infrastructure in NativeScript refer to [this article]({%slug plugins-infrastructure%}).
 
-  public set myProperty(value: number) {
-    this._myProperty = value;
-  }
+## Basic UI Plugin Structure
+
+The first file `index.d.ts` defines the public API of the control (in our case MyButton).
+
+_index.d.ts_
+```
+export * from "./my-button";
+```
+For better code readability the exported logic comes from `my-button.d.ts`
+It has dual purpose: as a contract when implementing the public API and to give intellisense when "my-button" is used in some editors/IDEs.
+
+_my-button.d.ts_
+```
+import { View, Style, Property, CssProperty, EventData } from "tns-core-modules/ui/core/view";
+
+export const textProperty: Property<MyButton, string>;
+export const myOpacityProperty: CssProperty<Style, number>;
+
+export class MyButton extends View {
+    // static field used from component-builder module to find events on controls.
+    static tapEvent: string; 
+
+    // Defines the text property.
+    text: string;
+
+    // Overload 'on' method so that it provides intellisense for 'tap' event.
+    on(event: "tap", callback: (args: EventData) => void, thisArg?: any);
+
+    // Needed when 'on' method is overriden.
+    on(eventNames: string, callback: (data: EventData) => void, thisArg?: any);
 }
 ```
 
-The TypeScript transpiler is run by a grunt script.
+In the file above, we declare that our control will have `text` and `myOpacity` properties. We also provide intellisense when `myButton.on` method is called so that it is known that a tap event is exposed. There is also definition of two properties - `text: Property` and `myOpacity: CssProperty`.
+Further info on what `Property` and `CssProperty` means is covered [here](#property-class).    
 
-NativeScript features two types of properties: **dependency properties** and **style properties**. Each type is described in the next sections.
+The next file is `my-button-base.ts`. In the base file we define all common fields, properties and methods that are applicable for both Android and iOS. At the top of the file we declare our new properties `text: Property` and `myOpacity: CssProperty`.
 
-
-## Dependency properties
-
-This sections includes these topics:
-
-###<a id = "intro1"></a>Introduction
-
-Dependency properties provide valuable features that simplify the creation of a rich User Interface (UI), including: 
-
-* Memory optimization: The creation of a rich custom UI control is bound to creating a huge number of properties, most of which are used with default values. With the traditional approach you end up with a **private field** for every property. With dependency properties, you only store the instance properties that you modified. The default values are stored within the dependency property. Additionally, dependency properties are declared statically outside the class, which further helps optimize the memory footprint.
-* Value validation: Dependency properties offer business logic validation. It is implemented as a dedicated validation callback function that takes `newValue` as parameter and returns true or false if the value is valid or not, respectively.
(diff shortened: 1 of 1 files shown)


core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
-  }
-
-  public set myProperty(value: number) {
-    this._myProperty = value;
-  }
-}
-```
-
-The TypeScript transpiler is run by a grunt script.
+Being a TypeScript framework, NativeScript uses TypeScript properties. After transpilation, these result in ECMAScript v.5 compliant JavaScript with setter and getter methods to support working with class members, thus ensuring readable and manageable code. This article will guide you through the process of creating basic UI plugin using the property system. All the code is in TypeScript and requires NativeScript 3.x.x or newer version.
```
hshristov, line 25: This sounds out of context to me. What getter & setter methods? What readable & manageable code?
Better remove that fluff :)

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
 
-function onMyPropertyChanged(eventData: dependencyObservable.PropertyChangeData) {
-  var myClassInstance = <MyClass>eventData.object;
-  var value = eventData.newValue;
+In the file above, we declare that our control will have `text` and `myOpacity` properties. We also provide intellisense when `myButton.on` method is called so that it is known that a tap event is exposed. There is also definition of two properties - `text: Property` and `myOpacity: CssProperty`.
+Further info on what `Property` and `CssProperty` means is covered [here](#property-class).    
 
-  console.log("myProperty of the object " + myClassInstance.toString() + " changed with " + value);
-}
+The next file is `my-button-base.ts`. In the base file we define all common fields, properties and methods that are applicable for both Android and iOS. At the top of the file we declare our new properties `text: Property` and `myOpacity: CssProperty`.
```
radeva, line 86: I would suggest something like "In `my-button-base.ts` we define..." instead of "The next file is `my-button-base.ts`. In the base file"

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
+`textProperty: Property<MyButtonBase, string>` - here the owning type is MyButtonBase meaning that this property will be defined on instances of MyButtonBase. The type of the property is `string` so it will accept any text. 
 
-###<a id = "setting"></a>Setting a style property
+If the type of the property not `string` we will probably need to specify `valueConverter` and `equalityComparer`. The `valueConverter` will be called if a string value is set to your property (for example from xml or css) and there you will have to convert that string to meaningful value if possible or throw exception if you can't. If `equalityComparer` is specified it will be called everytime a value is set to a property. There you can compare current and new value for equality. For example if your property is of type `Color` you can use `Color.equals` as `equalityComparer` function so even if new instance of `Color` is set the comparer will return `false` if current color and new color have the same `argb` value.
```
hshristov, line 150: Missing **is**
If the type of the property **isn't**

hshristov: to your property
may better as **to this property**

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
+├── my-button.android.ts
+├── package.json
+├── references.d.ts
+├── tsconfig.json
+└── platforms/
+    ├── android/
+    │   ├── include.gradle
+    └── ios/
+        └── Info.plist
+```
```
radeva, line 47: I would suggest here to keep the naming of the files consistent with the plugin seed. We discussed it in the morning with @vakrilov and decided that the current seed naming is ok. It's not a big deal but I think it will help the users to map the example with the seed. So it's up to you.

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
 
-Setting a style property is similar to setting a regular property but you use the nested `style` object (that is a property of the View class, which means that every UI component has style).
+There is one more property in the `Property` constructor: `affectsLayout: boolean`. When set to `true` setting new value to this property will trigger a new layout pass. `textProperty` sets `affectsLayout: isIOS`. This means that this property will request new layout pass only for `ios`. This is done as performance optimization. `android` has an integrated layout system so most of the time it will invalidate it self when needed. Thus we skip one native call by defining `affectsLayout` as `true` only for `ios`. But `ios` doesn't have integrated layout system so if you know that your property could affect the layout you should specify it in the `Property` constructor.
```
hshristov, line 152: There is one more property
**There is one more option**

core-concepts/properties.md
radeva, line 150: "If the type of the property not `string`" -> If the type of the property is not `string` : a missing "is"

core-concepts/properties.md
hshristov, line 152: only for `ios`
only when executing on `ios`

hshristov, line 152: But `ios` doesn't have
**Because `ios` doesn't have**

core-concepts/properties.md
radeva, line 150: Nothing is mentioned about `valueChange`. Although it's self-explanatory maybe it would be nice to have some example about it also.

core-concepts/properties.md
hshristov: Because `ios` doesn't have integrated layout system if you know that this property could affect the layout you should specify it in the `Property` constructor.

hshristov (changes requested):

core-concepts/properties.md
tjvantoll: I find the transition from “NativeScript uses TypeScript properties” to “This article will guide you through the process of creating basic UI plugins” a little strange.

Are properties only applicable to UI elements? If so, this article should state that. Should readers only care about this article if they’re building user interface plugins? (If so, this article shouldn’t be considered a “core concept”, because only a handful of our users will find this information valuable.

core-concepts/properties.md
tjvantoll: > If the type of the property not `string` we will probably need to specify `valueConverter` and `equalityComparer`.

“Probably” is not a good word to use in documentation 😄 

The docs should spell out exactly when you need `valueConverter` and `equalityComparer`.

core-concepts/properties.md
tjvantoll: I think I’ve asked this one before: but given the way this paragraph is written I wonder why we wouldn’t just make `affectsLayout: isIOS` the default. Everyone reading this will just do that because we tell them to.

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
-  }
-}
-```
-###<a id = "changedCallback"></a>Adding a changed callback
+    // Defines the text property.
+    text: string;
 
-__Example 3__ demonstrates how to implement the functionality of changing a notification. It adds a callback function, `onMyPropertyChanged`, that prints a message about a change in a property.
+    // Overload 'on' method so that it provides intellisense for 'tap' event.
+    on(event: "tap", callback: (args: EventData) => void, thisArg?: any);
```
DimitarTachev, line 76: You could also think about using TapEventData instead of the default EventData. The example will make more sense.

tjvantoll (commented):
I left this in a specific comment, but I thought I’d mention this overall too: this feels like an article on how to build UI plugins that happens to be titled “Properties” and put in the “Core Concepts” folder.

Is there a reason that all NativeScript developers need to know about properties? If yes, I think we should start the article with generic property information for non-plugin authors. If no, I think we should rename this article “Building UI Plugins” and put it in the Plugins folder.

core-concepts/properties.md
PetyaSotirova: I'm with TJ here - the content flow feels unnatural as if this article contains two separate topics squished together.

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
+├── package.json
+├── references.d.ts
+├── tsconfig.json
+└── platforms/
+    ├── android/
+    │   ├── include.gradle
+    └── ios/
+        └── Info.plist
+```
+> Note: For more details about the plugin infrastructure in NativeScript refer to [this article]({%slug plugins-infrastructure%}).
```
PetyaSotirova, line 48: What's the purpose of having this here if it is already available in the Plugins article?

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
 
-###<a id = "intro1"></a>Introduction
+## Basic UI Plugin Structure
 
-Dependency properties provide valuable features that simplify the creation of a rich User Interface (UI), including: 
+The first file `index.d.ts` defines the public API of the control (in our case MyButton).
```
PetyaSotirova, line 52: I'd drop "first" here - the 'index.d.ts file defines...

core-concepts/properties.md
PetyaSotirova: Additionally to what TJ said, I'd change the tense in this paragraph to present simple. The valueConverter IS called when the property is changed.

core-concepts/properties.md
```diff
@@ -10,515 +10,379 @@ environment: nativescript
...
+ 
+There are three important methods: 
+- `createNativeView` - you override this method, create and return your nativeView 
+- `initNativeView` - in this method you setup listeners/handlers to the nativeView 
+- `disposeNativeView` - in this method you clear the reference between nativeView and javascript object to avoid memory leaks as well as reset the native view to its initial state if you want to reuse that native view later.
+ 
+In this implementation we use singleton listener (for android - `clickListener`) and handler (for ios - `handler`) in order to reduce the need to instantiate native classes and to reduce memory usage. If possible it is recommended to use such techniques to reduce native calls.
```
PetyaSotirova, line 388: Most of the platform specific code above is documented. There are three important methods:
•	_createNativeView_: Override this method, create and return your nativeView.
•	_initNativeView_: Setup listeners/handlers to the nativeView.
•	_disposeNativeView_: Clear the reference between nativeView and javascript object in this method to avoid memory leaks. Reset the native view to its initial state if you want to reuse that native view later.

In Android, avoid access to native types in the root of the module (note that ClickListener is declared and implemented in a function which is called at runtime). This is specific for the V8 snapshot feature which is generated on a host machine where android runtime is not running. What is important is that if you access native types, methods, fields, namespaces, etc. at the root of your module (e.g. not in a function) your code won't be compatible with V8 snapshot feature. The easiest workaround is to wrap it in a function like in the above initializeClickListener function.

In this implementation, we use singleton listener (for android - clickListener) and handler (for ios - handler) in order to reduce the need to instantiate native classes and to reduce memory usage. If possible it is recommended to use such techniques to reduce native calls.

PetyaSotirova:
Not all title headings are with the proper casing, e.g. Platform - **S**pecific **C**ode.
Please use imperative mood wherever possible to avoid ambiguity. For example, this sentence:
"We also provide intellisense when myButton.on method is called so that it is known that a tap event is exposed."
Who is "we" - we who develop the framework or we who are going through this tutorial.

PetyaSotirova (changes requested):

core-concepts/properties.md
NickIliev: @tjvantoll, @PetyaSotirova perhaps the focus of this article should be "Creating NativeScript Plugins" and should replace the old UI plugins article? Let me know how you feel about it.

core-concepts/properties.md
NickIliev: @radeva meaning that the users have no need of Info.plist file for iOS when creating plugins via plugin-seed?

core-concepts/properties.md
NickIliev: In the plugins article, we have the detailed explanation of a NativeScript plugin. Here we are showing what the users will need to go step-by-step through the specific article - e.g. where to place their prerequisite files like `my-button.d.ts` or `my-button.android.ts`

NickIliev:
We have discussed the main purpose of this article and I will re-write it so it will be strictly Property System related and not structured like it is a tutorial for a UI plugin. Closing this PR in favor for upcoming new one.

Issue #796, opened 2017-06-05
Extending native apps with NativeScript + JS


PetyaSotirova:
Create articles and accompanying examples showing how to extend a native iOS and Android apps with NativeScript and JavaScript/TypeScript. The content available now only covers Angular: https://docs.nativescript.org/angular/integration-with-existing-ios-and-android-apps/extend-existing-android-app-with-ns-angular2.html.

- [ ] iOS
- [x] Android

PR #802, opened 2017-06-09
Revamp the extend-existing-android-apps article

petekanev:
Addresses https://github.com/NativeScript/docs/issues/796 - content is common for both Angular and JS NativeScript apps.


PetyaSotirova (changes requested):
Please add ReadMe to the sample app repository and cross-link with the article.
The article is set with `environment: angular` which will hide it from the core docs.

PetyaSotirova (approved):

PR #805, opened 2017-06-16
Rework of the plugin documentation to include new guides on using/building plugins

tjvantoll:
Since some of these diffs can be a bit hard to read, here’s basically what’s happening. The new proposed TOC of the “plugins” folder with this PR is:

```
Plugins/
    Using Plugins
    Building Plugins
    Building Plugins That Use CocoaPods
    Plugin Reference
```

[**Using Plugins**](https://github.com/tjvantoll/docs/blob/plugin-guides/plugins/plugins.md) is a new short article that documents the basics of finding and installing NativeScript plugins.

[**Building Plugins**](https://github.com/tjvantoll/docs/blob/plugin-guides/plugins/building-plugins.md) is a new not-short article that walks readers through the basics of creating a plugin from scratch.

**Building Plugins That Use CocoaPods**—This is an old article. I changed the name to fit the naming structure of the new articles.

**Plugin Reference** has the contents of the https://docs.nativescript.org/plugins/plugins, but has a new intro to tell readers that this is more advanced material. (IMO we would prefer readers to see the documentation on the plugin seed first.) There’s still some essential stuff in this article, so we want this page to remain on the docs.

Hopefully this structure makes it clear where other articles can fit in, for example Debugging Plugins, or Building User Interface Plugins.

Let me know if you have any comments or concerns. I will be out of the office next week so Fluent, so my apologies if my replies are a little late 😄


tjvantoll:
Squashed and merged with 081e0ab5f476f9d43edfd72d1dc65e60334c957f.

PR #816, opened 2017-06-23
Properly defined android requirements

OscarArnaizLopez:
Android requirement properly defined.

### DIFF
--- a/runtimes/android/requirements.md
+++ b/runtimes/android/requirements.md
@@ -9,7 +9,7 @@ position: 2
 The recommended ways to create NativeScript Applications is either through the [Command-Line Interface (CLI)](https://github.com/NativeScript/nativescript-cli) or through the [Telerik AppBuilder](http://www.telerik.com/appbuilder).
 
 # Supported API Levels
-The Android Runtime is built against [API level 17](http://developer.android.com/about/versions/android-4.2.html). The minimum supported compile SDK version is 22. Additionally, using the [metadata-generator](metadata/generator.md) tool you may generate your own metadata for any Android SDK available in the Android SDK Manager.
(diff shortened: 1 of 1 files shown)


runtimes/android/requirements.md
```diff
@@ -9,7 +9,7 @@ position: 2
 The recommended ways to create NativeScript Applications is either through the [Command-Line Interface (CLI)](https://github.com/NativeScript/nativescript-cli) or through the [Telerik AppBuilder](http://www.telerik.com/appbuilder).
 
 # Supported API Levels
-The Android Runtime is built against [API level 17](http://developer.android.com/about/versions/android-4.2.html). The minimum supported compile SDK version is 22. Additionally, using the [metadata-generator](metadata/generator.md) tool you may generate your own metadata for any Android SDK available in the Android SDK Manager.
+The Android Runtime is built against [API level 22](https://developer.android.com/about/versions/android-5.1.html). The minimum supported compile SDK version is [API level 17](http://developer.android.com/about/versions/android-4.2.html). Additionally, using the [metadata-generator](metadata/generator.md) tool you may generate your own metadata for any Android SDK available in the Android SDK Manager.
```
petekanev, line 12: The android runtime is in fact built using [compileSdk 23](https://github.com/NativeScript/android-runtime/blob/master/runtime/build.gradle#L199),  with support in mind of devices as low as API Level 10.

The core-modules however, that the technology depends on, and is the core of all that is NativeScript, can be ran on devices as low as API Level 17.

Now, when building applications, you can explicitly specify the `compileSdk` version you want to build your app against, and that needs to be 22+ at the time of writing, because there are certain components that can't compile if built against a lower compileSdk level, hence the original text above.

petekanev (changes requested):
I do not agree with the changes for reasons explained in my comment. Feel free to suggest a better wording however, if the current one is confusing.

Thank you!

petekanev:
Hey @OscarLopezArnaiz, would you be willing to make changes to those already proposed ?

PR #828, opened 2017-07-04
Tbozhikov/ui plugin custom

tbozhikov:
- new article about Building UI Plugins Using Custom Components
- fixes in the article Building UI Plugins Using Composite Components


plugins/ui-plugin-custom.md
```diff
@@ -0,0 +1,310 @@
...
+
+# Building UI Plugins using Custom Components
+
+Whenever needed UI can be shown by a plugin just by exposing a custom component, e.g. some platform-specific functionality that renders UI itself. To demonstrate that, this article explains how to create a simple button plugin.
+
+Contents:
+
+* [Prerequisites](#prerequisites)
+* [Bootstrap Your Plugin](#bootstrap-your-plugin)
+* [Add the UI bits: common code](#add-the-ui-bits-common-code)
```
radeva: you can use h3 for "common-code" and "specific code" i.e.:
## Add the UI Bits
### Common Code
### Platform Specific Code

plugins/ui-plugin-custom.md
```diff
@@ -0,0 +1,310 @@
...
+* [Add the UI bits: platform-specific code](#add-the-ui-bits-platform-specific-code)
+* [Make Your Plugin Angular Compatible](#make-your-plugin-angular-compatible)
+
+## Prerequisites
+
+The article contains information applicable to apps built on NativeScript 3.x.x or newer version
```
radeva: I think the correct is "built with" instead of "built on"

plugins/ui-plugin-custom.md
```diff
@@ -0,0 +1,310 @@
...
+
+```
+    <ui:MyButton text="MyButton1" tap="onTap" />
+```
+
+This can be accomplished by wrapping the platform-specific buttons (iOS's UIButton and Android's android.widget.Button) and expose it from a common MyButton class.   
+
+You can implement this by creating four files:
+
+- **my-button.d.ts** - to hold the declarations of MyButton class, its properties "text" and "myOpacity", enables auto-complete in some IDEs.
```
radeva: keep the same formatting in the explanation. below "holds" is used which I think is better.

plugins/ui-plugin-custom.md
```diff
@@ -0,0 +1,310 @@
...
+    <ui:MyButton text="MyButton1" tap="onTap" />
+```
+
+This can be accomplished by wrapping the platform-specific buttons (iOS's UIButton and Android's android.widget.Button) and expose it from a common MyButton class.   
+
+You can implement this by creating four files:
+
+- **my-button.d.ts** - to hold the declarations of MyButton class, its properties "text" and "myOpacity", enables auto-complete in some IDEs. 
+- **my-button.common.ts** - contains the logic accessible from the apps.
+- **my-button.ios.ts** - holds the iOS-specific logic for creation the native view (UIButton)
```
radeva: "for creation the native view" -> "for creation of the native view"

plugins/ui-plugin-custom.md
```diff
@@ -0,0 +1,310 @@
...
+
+    // set opacity to the native view.
+    [myOpacityProperty.setNative](value: number) {
+        return this.nativeView.setAlpha(value);
+    }
+}
+```
+
+NOTE: In Android, avoid access to native types in the root of the module (note that ClickListener is declared and implemented in a function which is called at runtime). This is specific for the V8 snapshot feature which is generated on a host machine where android runtime is not running. What is important is that if you access native types, methods, fields, namespaces, etc. at the root of your module (e.g. not in a function) your code won't be compatible with V8 snapshot feature. The easiest workaround is to wrap it in a function like in the above `initializeClickListener` function.
+
```
radeva: Is there an article or blog post about snapshots? If there is, you can link it here.

plugins/ui-plugin-custom.md
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    // set opacity to the native view.
+    [myOpacityProperty.setNative](value: number) {
+        return this.nativeView.alpha = value;
+    }
+}
+``` 
+
+In the a.m. implementations we use singleton listener (for Android - `clickListener`) and handler (for iOS - `handler`) in order to reduce the need to instantiate native classes and to reduce memory usage. If possible it is recommended to use such techniques to reduce native calls.
```
radeva: What is "a.m." ?

radeva (changes requested):

plugins/ui-plugin-custom.md
tbozhikov: "a.m." = above mentioned
:)

radeva (approved):

PR #830, opened 2017-07-05
Add more documentation for V8 Heap Snapshot feature

ivanbuhov:
Add more documentation for:
 * `tnsJavaClassesOptions` configuration (FYI: @sis0k0 )
 * `global.__snapshot`, `global.__snapshotEnabled` flags
 * `genrate-android-snapshot` npm script


best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
...ed by the static binding generator and JavaScript counterpart, created when the [JavaScript extending](https://docs.nativescript.org/runtimes/android/generator/extend-class-interface#classes) is evaluated. We need both parts to be loaded before instantiating the class. The native Java definition is compiled with the app, so the class is loaded by the JVM without requiring any additional work. Since the JavaScript counterpart is created by [the extend function](https://docs.nativescript.org/runtimes/android/generator/how-extend-works#the-extend-function) we should make sure that the it is evaluated before using the successor class. This is not a concern, when the class is instantiated from JavaScript, because its constructor is created by the `extend` function call, which means that there is no way to refer to the extended class without having the `extend` call executed beforehand. However, there are classes defined in JavaScript that are instantiated from native code. For example, the `NativeScriptActivity` class defined in `ui/frame/activity` is instantiated on app startup by the system. When bundling is disabled, such cases are handled by the Android runtime in the following way:
```
PetyaSotirova: Every Java class successor declared in NativeScript, consists of two parts - a native Java part generated by the static binding generator and a JavaScript counterpart,...
Who is "we" in the next sentence? Please rephrase as it is a bit confusing.

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+    1. The static binding generator saves in the class metadata the path to the file containing its JavaScript implementation.
```
PetyaSotirova: Reverse the word order - saves the path in the class...

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+    ```
+    plugins.push(new nsWebpack.NativeScriptSnapshotPlugin({
+        ...
+        tnsJavaClassesOptions: { modules: ["path/to/file1.js", "path-to-module", "path/to/other-script"] }
+        ...
+    }));
+    ```
+    If you are a plugin author and your plugin contains such modules that have to be listed here, you can [specify it in your plugin's `package.json` file](https://github.com/NativeScript/NativeScript/commit/67a8144a3439f3a71f3cb3fb0715289c66c014d5). This gives the opportunity to the plugin's clients to add all needed paths in their `tns-java-classes.js` file only by specifying the name of the containing package:
```
PetyaSotirova: Either "contains a module" or "specify them"

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+    ```
+    plugins.push(new nsWebpack.NativeScriptSnapshotPlugin({
+        ...
+        tnsJavaClassesOptions: { packages: ["my-package" ] }
+        ...
+    }));
+    ```
+#### Checking if snapshot is enabled
+If you have logic that you want (not) to be executed only in snapshotted context you can check the `global.__snapshot` flag. Its value is `true` only if the current execution happens in the snapshotted context, triggered by `mksnapshot` tool. Once the app is deployed on the device, the value of the flag is changed to `false`. There is also `global.__snapshotEnabled` flag. Its only difference compared to `global.__snapshot` is that its value is `true` in both snapshotted and runtime contexts, given that snapshot generation is enabled.
```
PetyaSotirova: If you want to toggle whether specific logic is executed only in snapshot context, you can...

PetyaSotirova (changes requested):

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+        if (!global.__snapshot) {
+            console.log("The current execution is happening in runtime context when we have all {N} APIs available, including console.log, so this line of code won't fail.");
+        }
+        console.log("This will fail if logMessage is called in snapshotted context because console.log is not available there.");
+    }
+}
+```
+
+#### Using snapshot without Webpack
+Using snapshot without Webpack is not a common scenario. Packing all scripts into one fat bundle, in advance, is what makes the snapshot so beneficial. Generating snapshot for a single script only will rarely have some notable effect on the app performance. However you can still do it. The `nativescript-dev-webpack` plugin adds `genrate-android-snapshot` npm script in the app's `package.json` file. The command will check if `__snapshot.js` file exists in the app root folder and will generate and install blob files using the file as input.
```
sis0k0: `genrate-android-snapshot` -> `generate-android-snapshot`

sis0k0: Maybe, it will be better to start with how to do it and then explain that it's not a common scenario.

best-practices/bundling-with-webpack.md
sis0k0: - Add link to the SBG docs -
 https://docs.nativescript.org/runtimes/android/generator/overview
- `we should make sure that the it is evaluated` -> `we should make sure that it is evaluated `

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+    1. The static binding generator saves in the class metadata the path to the file containing its JavaScript implementation.
+    2. Before instantiating the class, the runtime loads and executes the module at the path saved in the metadata, assuming that this will evaluate the class extending code.
+    This doesn't work in bundling scenario, because the saved path in the metadata always points to the bundle. However, executing the bundle script will not achieve the desired result. Therefore, we add `require` calls to such modules in our default `vendor-platform.android.js`:
```
sis0k0: `However, executing the bundle script will not achieve the desired result.`
Shall we explain why?

sis0k0: `vendor-platform.android.js` -> `app/vendor-platform.android`

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+        /*
+        In case snapshot generation is enabled these modules will get into the bundle but will not be required/evaluated.
+        The snapshot webpack plugin will add them to the tns-java-classes.js bundle file. This way, they will be evaluated on app start as early as possible.
+        */
+        require("ui/frame");
+        require("ui/frame/activity");
+    }
+    ```
+    However, when we add the V8 heap snapshot in the whole story, we are not allowed to extend native classes in snapshotted context. Therefore, we always include `ui/frame` and `ui/frame/activity` in the bundle but when snapshot is enabled they are not evaluated. Here is where the `tns-java-classes.js` files comes into play. Just after initialization, the Android runtime [will evaluate the script at `app/tns-java-classes.js` path if such exists](https://github.com/NativeScript/android-runtime/commit/d13189e4206b374142dc61d309d5aa708fb8095f). The snapshot generator creates such file and populates it with `require` calls to a user defined list of paths. This way, they are explicitly evaluated on app startup. You can add paths to be `require`-d in `tns-java-classes.js` like this:
```
sis0k0: ``the `tns-java-classes.js` files`` -> ``the `tns-java-classes.js` file``

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+    ```
+    plugins.push(new nsWebpack.NativeScriptSnapshotPlugin({
```
sis0k0: drop plugins.push

best-practices/bundling-with-webpack.md
sis0k0: Let's inline the example configuration in the docs instead of link to a commit.

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,52 @@ if (env.snapshot) {
...
+        ...
+        tnsJavaClassesOptions: { modules: ["path/to/file1.js", "path-to-module", "path/to/other-script"] }
+        ...
+    }));
+    ```
+    If you are a plugin author and your plugin contains such modules that have to be listed here, you can [specify it in your plugin's `package.json` file](https://github.com/NativeScript/NativeScript/commit/67a8144a3439f3a71f3cb3fb0715289c66c014d5). This gives the opportunity to the plugin's clients to add all needed paths in their `tns-java-classes.js` file only by specifying the name of the containing package:
+    ```
+    plugins.push(new nsWebpack.NativeScriptSnapshotPlugin({
```
sis0k0: drop plugins.push

best-practices/bundling-with-webpack.md
sis0k0: Explain what's snapshotted context.

sis0k0 (changes requested):

best-practices/bundling-with-webpack.md
```diff
@@ -200,7 +200,7 @@ $ npm run build-ios-bundle --uglify -- --release --forDevice --teamId TEAM_ID
...
+The webpack configuration also includes the [`NativeScriptSnapshotPlugin`](https://github.com/NativeScript/nativescript-dev-webpack/blob/master/plugins/NativeScriptSnapshotPlugin.js). The plugin loads a single webpack bundle in an empty V8 context, aka snapshtted context, and after its execution, captures a snapshot of the produced V8 heap and saves it in a blob file. Next the blob file is included in the apk bundle and [is loaded by the Android Runtime](https://docs.nativescript.org/runtimes/android/advanced-topics/V8-heap-snapshots) on app initialization. This will obviate the need for loading, parsing and executing the script on app startup which can drastically decrease the starting time.
```
sis0k0: typo
`snapshtted` -> `snapshotted`

PetyaSotirova (approved):

best-practices/bundling-with-webpack.md
```diff
@@ -244,6 +244,62 @@ if (env.snapshot) {
...
+        if (!global.__snapshot) {
+            console.log("The current execution is happening in runtime context when we have all {N} APIs available, including console.log, so this line of code won't fail.");
+        }
+        console.log("This will fail if logMessage is called in snapshotted context because console.log is not available there.");
+    }
+}
+```
+
+#### Using snapshot without Webpack
+The `nativescript-dev-webpack` plugin adds `generate-android-snapshot` npm script in the app's `package.json` file. The command will check if `__snapshot.js` file exists in the app root folder and will generate and install blob files using the file as input. Make sure you have prepared the android platform beforehand, because a subsequent prepare will clear some installed artefacts. This is how V8 heap snapshot can be generated from a single file, without using Webpack bundling. However, generating snapshot from something different than a Webpack bundle is not a common scenario. Packing all scripts into one fat bundle, in advance, is what makes the snapshot so beneficial. Generating snapshot for a single script only will rarely have some notable effect on the app performance.
```
sis0k0: `into one fat bundle, in advance`, drop the comma?

sis0k0 (approved):

PR #832, opened 2017-07-08
Update ng-chapter-3.md

developer82:
"Exercise: Use an Angular service" was missing a step to import the Http module, causing an error for those who follow the tutorial step by step and are not familiar with Angular

### DIFF
--- a/tutorial/ng-chapter-3.md
+++ b/tutorial/ng-chapter-3.md
@@ -448,6 +448,37 @@ export class UserService {
 }
 ```
 
+Notice that the code above uses a new Http module, that we first need to introduce, just like other modules we have used before in this toturial.
+Open `app/app.module.ts`, and import the `NativeScriptHttpModule` by adding the following import line at the top of the file:
+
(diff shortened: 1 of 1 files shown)


tjvantoll:
Hi @developer82,

Thanks so much for this, but there’s a section right below on “Exercise: Create an account” that has you add this import to make that example work. I’m curious, did you eventually find that section and get things working?

I’m thinking what might be best here is a quick warning that tells the reader that the registration call won’t work _yet_, and that you’ll need to wait until you complete the next exercise. What do you think?

developer82:
@tjvantoll Hi, I found that section, but only after I had fixed the problem by searching google, and read a post on StackOverflow aimed specially for NS developers.
I think this section should be introduced earlier, since developers probably won't understand why things aren't working before moving on to the other section.

tjvantoll:
We’re incorporating this feedback into the next version of these tutorials, which changes how this section works a bit. Thanks again @developer82!

PR #840, opened 2017-07-12
Supporting Angular in UI Plugins

DimitarTachev:


plugins/angular-plugin.md
```diff
@@ -0,0 +1,84 @@
+---
+title: Supporting Angular in UI Plugins
+description: Supporting Angular in UI Plugins
+position: 10
+slug: supporting-angular-in-ui-plugins
+publish: false
```
PetyaSotirova: This will add the article to the site but exclude it from the TOC tree. I assume you wanted to add 'environment: angular' instead.

plugins/angular-plugin.md
```diff
@@ -0,0 +1,84 @@
...
+---
+
+# Supporting Angular in UI Plugins
+
+In order to make your UI plugin Angular compatible you need to create a simple Angular wrapper. Before you continue, make sure you have covered the topic about [Building UI Plugins using Custom Components]({% slug building-ui-plugins-custom-components %}) or the one for [Building UI Plugins using Composite Components]({% slug building-ui-plugins-composite-components %}).
+
+
+##  Angular Wrapper
+
+Having your UI plugin developed successfully you could easily make it Angular compatible following the steps below:
```
PetyaSotirova: Once your UI plugin is ready, you can easily ...

PetyaSotirova (commented):
Both of these articles are at position 10, consider changing.

PR #849, opened 2017-07-22
Linux installation docs update

jorotenev:
Proposing to add an extra step in the Linux installation step - to enable the Developer mode + USB debugging on the emulated device. Without this step `tns run android` was failing with ```Cannot run your app in the native emulator. Increase the timeout of the operation with the --timeout option or try to restart your adb server with 'adb kill-server' command. Alternatively, run the Android Virtual Device manager and increase the allocated RAM for the virtual device."```

Enabling the developer mode + USB debugging solved the issue on both Windows and Linux for me.


etabakov (approved):

etabakov:
Thanks for the suggestion! This PR was outdated, so I merged your commit in a separate PR.

PR #862, opened 2017-08-07
Fix two links to other docs pages

airandfingers:
These two links point to the top-level page http://docs.nativescript.org/ui/basics#user-interface-components instead of the specific sections they describe.

This change makes the links point to specific sections - the list of components, and the custom components section.


etabakov:
Thanks for the suggestion @airandfingers!

PR #871, opened 2017-08-09
Update introduction.md

rodnolan:
The last part of this sentence is confusing: Use Angular, TypeScript or modern JavaScript to get truly native UI and performance "while sharing skills and code with the web". Why would the web want me to share my skills and code with it?


etabakov:
Thanks for the feedback @rodnolan. I understand where the confusion comes from. However, I think it's good to keep the point that you can reuse your code and skill from the web development. I will try to rephrase it in a way that sounds better.

PR #879, opened 2017-08-18
Removed duplicate topic getApplication method

lshens:
The topic is in the Android - Platform specific methods section


etabakov:
Thanks for pointing out to this issue. Actually there are two different methods: `utilsModule.ad.getApplication()` and `utilsModule.ad.getApplicationContext()`. I will change the title of second paragraph accordingly.

PR #880, opened 2017-08-21
doc: initial isLoading value should be true

jurgentreep:
The loading indicator doesn't show up in the app because the initial value of isLoading is set to false.


tjvantoll:
Sorry for the late reply here @jurgentreep. Merged with 5a5b9084fb90f159a6498c8db559602f8c279855. Thanks!

PR #885, opened 2017-08-24
Update typescript examples in the two-way bindings page

jalkoby:
Make typescript examples use typescript features


etabakov (approved):
Thanks for the improvement @jalkoby

PR #908, opened 2017-09-17
Update for 3.0 border style

saknarak:
https://www.nativescript.org/blog/per-side-borders-in-nativescript-css


etabakov (approved):
@saknarak Thanks for taking the time improve our documentation! Good catch.

PR #911, opened 2017-09-19
Fixes on Hierachical Selector section.

mrcastegt:
The text should go after the example of **A direct child selector**.


etabakov (approved):
Thanks for your contribution, @mrcaste!

PR #919, opened 2017-09-26
Added line-height css property

000panther:
Support for the line-height css property was enabled in https://github.com/NativeScript/NativeScript/issues/1664 - but there is no documentation yet.


etabakov (approved):
Many thanks for your contribution @000panther!

PR #922, opened 2017-09-27
Add a warning so users realize embedding is not officially supported

tjvantoll:


atanasovg:
Thanks, TJ. This clears our message and sets proper expectations.

Issue #920, opened 2017-09-26
wget: unable to resolve host address ‘nsbuild01.telerik.com‘


engineervix:
Howdy folks, I have been trying to setup the NativeScript docs on a local machine, [as described here](https://github.com/NativeScript/docs#local-setup). Everything went fine, until I ran the `./build.sh` script. I get the following output:

```
+ ENV=stable
+ . ./dependencies.sh
+ PACKAGE_VERSION=0.0.0
+ export TARGET_DIR=dist
+ TARGET_DIR=dist
+ rm -rf dist
+ mkdir -p dist
+ get_dependencies
+ rm -rf '*.tar.gz'
+ rm -rf _dependencies
+ mkdir -p _dependencies
+ cd _dependencies
+ get_artifacts sdk-examples-docs-stable 1
+ BUILD_NAME=sdk-examples-docs-stable
+ shift
+ ((  1  ))
+ ARTIFACT_ID=1
+ shift
+ JENKINS_PREFIX=http://nsbuild01.telerik.com:8080/build/view/Docs/job/
+ JENKINS_SUFFIX=/lastSuccessfulBuild/deployedArtifacts/download/artifact.1/
+ URL=http://nsbuild01.telerik.com:8080/build/view/Docs/job/sdk-examples-docs-stable/lastSuccessfulBuild/deployedArtifacts/download/artifact.1/
+ wget --clobber --content-disposition http://nsbuild01.telerik.com:8080/build/view/Docs/job/sdk-examples-docs-stable/lastSuccessfulBuild/deployedArtifacts/download/artifact.1/
--2017-09-26 17:50:21--  http://nsbuild01.telerik.com:8080/build/view/Docs/job/sdk-examples-docs-stable/lastSuccessfulBuild/deployedArtifacts/download/artifact.1/
Resolving nsbuild01.telerik.com (nsbuild01.telerik.com)... failed: Name or service not known.
wget: unable to resolve host address ‘nsbuild01.telerik.com’
+ echo 'URL: http://nsbuild01.telerik.com:8080/build/view/Docs/job/sdk-examples-docs-stable/lastSuccessfulBuild/deployedArtifacts/download/artifact.1/'
URL: http://nsbuild01.telerik.com:8080/build/view/Docs/job/sdk-examples-docs-stable/lastSuccessfulBuild/deployedArtifacts/download/artifact.1/
+ ((  0  ))
+ get_artifacts nativescript-angular-docs-stable 1
[...]

PR #928, opened 2017-10-02
Add --skipDependencies flag that bypasses the download of artefacts ...

etabakov:
… from internal server


etabakov:
Temporary workaround for [#920](https://github.com/NativeScript/docs/issues/920)

PR #930, opened 2017-10-04
Linux installation setup

etabakov:


etabakov:
Credits to @jorotenev for the suggestion!

PR #931, opened 2017-10-06
let JDK 8 link lead to jdk 8 not to latest java

Plamen5kov:
currently nativescript android doesn't work with java 9, so let's change the link to point to the real JDK 8 download page


etabakov (approved):
Good catch. Thanks!

PR #934, opened 2017-10-09
update article to include a more natural way of debugging

Plamen5kov:
Update the getting started article to show the normal debugging flow, besides console log.

### DIFF
--- a/docs/tutorial/ng-chapter-1.md
+++ b/docs/tutorial/ng-chapter-1.md
@@ -184,6 +184,8 @@ The default way of debugging a NativeScript application is by using the integrat
 **Visual Studio Code**
 One of the preffered ways of debugging a NativeScript application is with the [NativeScript extension for Visual Studio Code](http://docs.nativescript.org/tooling/debugging#debugging-with-visual-studio-code).
 
+You’ll want to check out the Chrome Dev Tools and the Visual Studio debugger as you have the need to step through your JavaScript code or TypeScript code, but for the moment let’s look at how to do some simple logging in your apps.
+
 <h4 class="exercise-start">
     <b>Exercise</b>: Console logging in NativeScript
 </h4>
--- a/docs/tutorial/ng-chapter-1.md
+++ b/docs/tutorial/ng-chapter-1.md
@@ -182,7 +182,7 @@ There are three main paths you can take when you want to debug a NativeScript ap
 The default way of debugging a NativeScript application is by using the integration with Chrome. You can find more information on the actual steps [here](http://docs.nativescript.org/tooling/debugging).
 
 **Visual Studio Code**
-One of the preffered ways of debugging a NativeScript application is with the [NativeScript extension for Visual Studio Code](http://docs.nativescript.org/tooling/debugging#debugging-with-visual-studio-code).
+One of the prefered ways of debugging a NativeScript application is with the [NativeScript extension for Visual Studio Code](http://docs.nativescript.org/tooling/debugging#debugging-with-visual-studio-code).
 
 You’ll want to check out the Chrome Dev Tools and the Visual Studio debugger as you have the need to step through your JavaScript code or TypeScript code, but for the moment let’s look at how to do some simple logging in your apps.
 
--- a/docs/tutorial/ng-chapter-1.md
+++ b/docs/tutorial/ng-chapter-1.md
@@ -184,8 +184,6 @@ The default way of debugging a NativeScript application is by using the integrat
 **Visual Studio Code**
 One of the preffered ways of debugging a NativeScript application is with the [NativeScript extension for Visual Studio Code](http://docs.nativescript.org/tooling/debugging#debugging-with-visual-studio-code).
 
-**ConsoleLog**
-
 <h4 class="exercise-start">
     <b>Exercise</b>: Console logging in NativeScript
 </h4>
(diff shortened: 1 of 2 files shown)


tsonevn (approved):

etabakov (changes requested):
I think this section is getting messy. We have three ways to debug an application and we are showcasing the most primitive and least preferable one. Isn't it better to change the title of the current one to "Logging" and create a new section for debugging with Chrome Dev Tools?

@tjvantoll What do you think about this?

docs/tutorial/ng-chapter-1.md
```diff
@@ -176,6 +176,14 @@ Overall, the NativeScript CLI’s livesync process makes native iOS and Android
 
 You now know how to create and run NativeScript apps. Your next step is learning how to debug your apps when things go wrong. Let’s look at how debugging works in NativeScript.
 
+There are three main paths you can take when you want to debug a NativeScript application:
+
+**Chrome**
+The default way of debugging a NativeScript application is by using the integration with Chrome. You can find more information on the actual steps [here](http://docs.nativescript.org/tooling/debugging).
+
+**Visual Studio Code**
+One of the preffered ways of debugging a NativeScript application is with the [NativeScript extension for Visual Studio Code](http://docs.nativescript.org/tooling/debugging#debugging-with-visual-studio-code).
```
tjvantoll: preferred

docs/tutorial/ng-chapter-1.md
```diff
@@ -176,6 +176,14 @@ Overall, the NativeScript CLI’s livesync process makes native iOS and Android
 
 You now know how to create and run NativeScript apps. Your next step is learning how to debug your apps when things go wrong. Let’s look at how debugging works in NativeScript.
 
+There are three main paths you can take when you want to debug a NativeScript application:
+
+**Chrome**
+The default way of debugging a NativeScript application is by using the integration with Chrome. You can find more information on the actual steps [here](http://docs.nativescript.org/tooling/debugging).
```
tjvantoll, line 182: I worry that you set readers up for failure by just linking to that page and not walking them through exactly how to use the Chrome Dev Tools. For example, the Chrome Dev Tools only work in iOS if you use the `--chrome` flag, and that fact is barely mentioned.

tjvantoll: I think this will be ok if you mention it after the `console.log()` example though, because then it’s kind of an aside for readers to check out when they’re ready.

tjvantoll (commented):
Long term I think this tutorial should actually walk the user through running the appropriate commands and teach them the Dev Tools with examples, as @etabakov suggests. The next iteration of the tutorial that Jen & I are thinking through now will almost certainly do that.

Until then I don’t mind mentioning these tools, but I’d kind of like this PR’s wording to be put _after_ the `console.log` example—just because I think it’ll flow better. Right now it’s a bit weird to say, “here are the best two ways to debug your apps but we’re not going to use them”.

`console.log()` is basic, _but_ it’s something the reader can do without installing something new, and without having to stop their app from running with `Ctrl` + `C` and learning a new `debug` command with a `--chrome` flag.

That’s my thoughts at least. If you just replace the sentences below with what you have here I think this should be good for now. (Also remember that there are two tutorials, so there’s a `chapter-1.md` file with this same wording.)

> “The console.log() and console.dir() functions are two of the many tools available when debugging your NativeScript applications. You might want to try out step debugging as you move into more advanced development”

docs/tutorial/ng-chapter-1.md
Plamen5kov: @tjvantoll I don't think we're setting anyone up for failure when we're pointing them to a specific article about debugging. The fact that ios debugging in chrome is working with the `--chrome` flag is one of the first things, mentioned in the article.
Furthermore, I don't think console.log is actual debugging. On this point, I think more like @etabakov : let's have a logging section, but not call console.log .. debugging.

Plamen5kov:
@tjvantoll I understand what you're saying, but here's what I see.
I go and look at the tutorial, which says that debugging === console.log and in the bottom, there's something about vs code extension, which I may or may not be using and not a word for real debugging on chrome and safari.
I suggest having a debugging section with a simple link to debugging article and logging section which suggests using console.log.

> “here are the best two ways to debug your apps but we’re not going to use them”

On this concern: we can just say for the sake of understanding the flow of the application, we can use console.log, which won't be misleading IMO.

tjvantoll:
@Plamen5kov Ok that makes sense. I think as long as you add some wording that makes what you have in this PR flow into the `console.log` example we’ll be good.

So I’d add something like...

_You’ll want to check out the Chrome Dev Tools and the Visual Studio debugger as you have the need to step through your JavaScript code or TypeScript code, but for the moment let’s look at how to do some simple logging in your apps._

docs/tutorial/ng-chapter-1.md
tjvantoll: I think having a different section eventually makes a lot of sense, but I’d be hesitant to add a section that just has two links.

You’ve convinced me that the wording you currently have is worth moving forward with. I just want a bit of text to transition from Chrome/VS Code into the `console.log` examples and 👍  from me.

tjvantoll:
@Plamen5kov It’s “preferred” (two _r_ s) 😉 

Update that and 👍  from me on merging this in.

Plamen5kov:
my bad

etabakov (approved):

Issue #938, opened 2017-10-13
Search button doesn't autofocus search bar


Plamen5kov:
__Problem__
* Open [docs](http://docs.nativescript.org/)
* click search button
* start writing

(expectation) text appears in search bar
(currently) nothing happens

__Solution__
On search button click, autofocus search bar

etabakov:
Uploaded on LIVE. Closing the issue.

PR #940, opened 2017-10-13
Fix search text field not being auto focused #938

etabakov:


Plamen5kov (commented):
I actually like the way it was done before, is there a way to get the old html template with the .search-input. And also, how can I build it locally and test the fix?

etabakov:
I'm not sure if we can control this template. However, I don't think the previous approach worked at all - you cannot select the element while it is hidden, which the old version is doing.

You can build locally and test following the steps in the README - https://github.com/NativeScript/docs#local-setup. Just checkout my branch of the docs repository and the rest is the same.

Plamen5kov:
Thank you, I'll test it out locally.

Plamen5kov:
I haven't been able to run the documentation on linux, so I can't test the fix on my machine. If you've tested it locally and it works and If you think the [attribute="value] way is sufficient fix, feel free to merge.

PR #941, opened 2017-10-13
Update README.md

Plamen5kov:


etabakov:
You can just leave `vagrant up --provider virtualbox`. `vagrant up` would use the default provider, which in most cases is set to virtualbox, but it's ok to be set explicitly.

Plamen5kov:
on linux it blows up, because it can't find the default config.vm.box

Plamen5kov:
Will update the README and open a new PR, when I figure out all the steps to build on fedora.

PR #942, opened 2017-10-13
snippets using moduleId to resolve relative paths (WebPack related)

NickIliev:
Related to [https://github.com/NativeScript/docs/issues/937](https://github.com/NativeScript/docs/issues/937)

Using `moduleiD : module.id` provides relative paths and also allows bundling with WebPack

@tjvantoll should we revise the source code in sample-Groceries as well? 
Let me know if you have some considerations against it.


tjvantoll:
This actually came up before in #762 which I completely dropped the ball on. I’m in favor of this change. Long term I would like an explanation of exactly what `moduleId: module.id` actually does, but for now let’s just get this in there to avoid potential problems.

etabakov (approved):

PR #943, opened 2017-10-13
Fixed method name in sample

celso-wo:


etabakov:
Thanks for the correction @celso-wo!

PR #944, opened 2017-10-14
Add open source argument to the vs-code evangelism

MonsieurMan:
In my opinion the most important point was missed by not stating it is open source.


etabakov:
Thanks for the addition @MonsieurMan!

PR #948, opened 2017-10-24
Update ng-chapter-3.md

wheresvic:
Fix the login page template path


tjvantoll:
Hey @victorparmar,

Thanks for the PR and sorry for the late reply here. The template path currently being used is intentional. Did it cause problems for you as you went through the tutorial? I just gave this a shot and I’m having any issues personally.

Thanks.

wheresvic:
Hi @tjvantoll,

I did have to set the path (I'm using Ubuntu) but maybe it's no problem on the other platforms. I don't have time to investigate again unfortunately so let's just close this MR and I'll raise another one if I come across it again, cheers!

PR #960, opened 2017-11-09
new article to provide guidance on architecture choice

rdlauer:
Dan requested this new doc based on an outline Sean Perkins came up with


etabakov:
Great article! My only concern is that we might set unrealistic expectations about our Vue support and readiness. Probably, some kind of disclaimer that the Vue support is still about to be on par with Angular would help. What do you think?

sipacate:
@etabakov  There is a notification in the article about Vue not being ready for prime time. What updates would you suggest? (here is a list of outstanding issues: https://github.com/rigor789/nativescript-vue/issues the expectation is they will all be addressed by end of year)

etabakov:
My point is that even if all the issues listed are resolved - Vue will still have significantly less amount of documentation, code samples, demo apps and plugins working with it. We also do not have the required testing infrastructure to be able to guarantee the same level of service as for Angular.
Probably, we can explain that the Vue integration is a community effort that is not driven by the core engineering team.

sipacate:
I think that is fine though. DevRel works on it and we also get lots of community contributions. It's open source thus, not everything requires the core team. Let's push this change. If there are community complaints, I'll handle them.

PR #963, opened 2017-11-15
Fixed example code

benjlin:
<ActionItem> is no longer a self closing tag, made it an explicit closing tag


etabakov (approved):
Thanks for your contribution!

PR #965, opened 2017-11-17
Update android-virtual-devices.md

TechnicalSoup:
Clarification of device selection in examples of tns run android.


etabakov (approved):
Thanks for your contribution!

PR #968, opened 2017-11-24
Update quick-setup.md

eyal7773:
in windows cmd need to be run as administrator.


etabakov (approved):
Thanks for the contribution!

PR #978, opened 2017-12-01
Update instructions for plugin link and unlink

radeva:
Since npm 5 `npm install <folder>` automatically creates sym link and it's not needed to explicitly link/unlink the plugin.


etabakov (approved):
Looks good to me!

PR #979, opened 2017-12-02
Fix Image tags missing closure tags

eyal7773:
Image tag in ns should be with </Image> clouse tag, and not as seen with self clouse.


etabakov (approved):
Thanks for your contribution!

PR #996, opened 2017-12-11
Added cross-links to {N} Core and {N} with Angular

toddanglin:
Since it may not be entirely obvious that these docs differ significantly for Core and Angular, adding cross-link in the Intro to help route people to the correct version of the styling docs page if they land here via Google search.


tjvantoll:
👍 

Ultimately we’ll want to address the UX of the whole site so this sort of thing isn’t necessary, but until we reach this magical utopia I think this will help.

One minor thing: I’m going to put this in a blockquote so it shows up as an aside to the main content. (It’ll be in one of those green boxes.)

tjvantoll:
Merged with 62d5b62fbddd59d91b9afb7c93cb89b33e5efada.

PR #997, opened 2017-12-11
Updated font and font-weight docs

toddanglin:
Fixed broken link to iOS fonts and added detail to font-weight property documenting ability to use numbers for font-weight

### DIFF
--- a/docs/ui/styling.md
+++ b/docs/ui/styling.md
@@ -99,7 +99,11 @@ page.css = "button { color: red }";
 page.css = "button { color: red }";
 ```
 
-After you have set the default CSS for the page, you can add to it using two methods: adding CSS from a string and adding CSS from a file.
+After you have set the default CSS for the page, you can add to it using two methods:
+
+1. Adding CSS from a string
+2. Adding CSS from a file
+
 {% endnativescript %}
 {% angular %}
 ### Component-specific CSS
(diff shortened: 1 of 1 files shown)


docs/ui/styling.md
```diff
@@ -480,7 +484,7 @@ This list of properties can be set in CSS or through the style property of each
 | `font-family`         | `fontFamily`          | Sets the font family of the matched view. |
 | `font-size`           | `fontSize`            | Sets the font size of the matched view (only supports device-independent units). |
 | `font-style`          | `fontStyle`           | Sets the font style of the matched view. Possible values: `italic`, `normal`. |
-| `font-weight`         | `fontWeight`          | Sets the font weight of the matched view Possible values: `bold`, `normal`. |
+| `font-weight`         | `fontWeight`          | Sets the font weight of the matched view Possible values: `bold`, `normal` OR `100`,`200`,`300`,`400`,`500`,`600`,`700`,`800`,`900`, where `400` is `normal` and `700` is `bold` |
```
tjvantoll: Well I learned something new here 😄

I ran this by the modules team. They suggested adding a quick note that says something like, “Note that not all fonts support all font weight variants.”. Thoughts on that?

toddanglin: Good point. I'll add that note.

I *think* this support for the alternative font-weight format came in one of the last few releases. I only discovered it because something related to these changes broke the way I was doing styles in one of my apps. <shrug>

tjvantoll:
Merged with b1d9179bb420af42ebd6aeba3c75169db72a429e and 76436cfbfcaacf1fdb9fc3384017faefda5a9819.

PR #1000, opened 2017-12-12
Added section discussing Platform Specific CSS

toddanglin:
Highlights techniques for applying and managing platform specific CSS in NativeScript with links to other relevant docs articles. Includes variations for Core and Angular.

### DIFF
--- a/docs/ui/styling.md
+++ b/docs/ui/styling.md
@@ -17,7 +17,6 @@ This article includes the following topics:
 * [Accessing NativeScript component properties with CSS](#accessing-nativescript-component-properties-with-css)
 * [Using Fonts](#using-fonts)
 * [Import External CSS](#import-external-css)
-* [CSS parser playground](http://iamdustan.com/reworkcss_ast_explorer/)
 
 ## Introduction
 
@@ -38,6 +37,8 @@ The CSS styles can be set on 3 different levels:
 * [Inline CSS](#inline-css): Applies directly to a UI view
 
 If there is CSS declared on different levels&mdash;all will be applied. The inline CSS will have the highest priority and the application CSS will have the lowest priority.
+
+It is also possible to apply [platform specific CSS](#platform-specific-css).
 
 ### Application-wide CSS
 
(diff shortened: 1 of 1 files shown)


docs/ui/styling.md
```diff
@@ -165,6 +166,56 @@ Similarly to HTML, CSS can be defined inline for a UI view in the XML markup:
...
+```CSS
+/* Import shared style rules */
+@import './home-common.css';
+
+/* Add iOS/Android specific rules (if any) */
+.mystyle { ... }
+```
+{% endangular %}
+
+At build time, NativeScript will automatically import the common styles and choose the correct iOS or Android stylesheet depending on the target build platform.
```
tjvantoll, line 217: A quick Playground link that does something really simple like make a label red on iOS and blue on Android might be nice, as there are a few moving pieces here.

docs/ui/styling.md
```diff
@@ -39,6 +38,8 @@ The CSS styles can be set on 3 different levels:
 
 If there is CSS declared on different levels&mdash;all will be applied. The inline CSS will have the highest priority and the application CSS will have the lowest priority.
 
+It is also possible to apply [platform specific CSS](#platform-specific-css).
```
tjvantoll: platform-specific when used as an adjective. That applies a few times in this PR 😄

docs/ui/styling.md
```diff
@@ -165,6 +166,56 @@ Similarly to HTML, CSS can be defined inline for a UI view in the XML markup:
...
+
+NativeScript conventions make it easy to apply platform specific CSS, either via separate stylesheets or via in-line declarations. For an overview of NativeScript's convention-based file name rules for targeting files at specific platforms and screen sizes, [refer to this article in the docs](https://docs.nativescript.org/core-concepts/navigation#supporting-multiple-screens). {% angular %} **NOTE:** If you are using Angular, file name rules do not work for targeting specific screen sizes or orientations. JavaScript is required to target styles at different screens at runtime. [See this article](https://dzone.com/articles/tablet-support-for-nativescript-with-angular) for an example of targeting styles at tablets with Angular.{% endangular %}
+
+There are 4 primary ways to target styles at iOS or Android:
+
+1. Platform specific stylesheets {% nativescript %}(`styles.ios.css`, `styles.android.css`){% endnativescript %}{% angular }(`styles.component.ios.css`, `styles.component.android.css`){% endangular %}
+2. Platform specific markup blocks (`<ios> ... </ios>`, `<android> ... </android>`)
+3. Platform specific attributes (`<Label ios:style="..." android:style="..."`)
```
tjvantoll: For 2 and 3 I feel like full examples might be nice. As is there might not be enough detail for users to see this note and know how to use these conventions. Something like...

2

``` XML
<ios>
  <Label text="iOS devices display this label"></Label>
</ios>
<android>
  <Label text="Android devices display this label"></Label>
</android>
```

3

``` XML
<Label ios:text="iOS devices display this label"></Label>
<Label android:text="Android devices display this label"></Label>
```

docs/ui/styling.md
```diff
@@ -165,6 +166,56 @@ Similarly to HTML, CSS can be defined inline for a UI view in the XML markup:
...
+4. Platform specific CSS rules (`.ios .mystyle { ... }`, `.android .mystyle { ... }`) _[*requires plugin](https://market.nativescript.org/plugins/nativescript-platform-css)_
+
+The most common and maintainable pattern for managing platform agnostic and platform specific styles in NativeScript is with multiple stylesheets and CSS imports.
```
tjvantoll: platform-agnostic 😄

docs/ui/styling.md
tjvantoll: Another idea—what if the link to the Playground came at the beginning of your explanation of this concept? It might help readers to have a reference to the final code to put your descriptions in context.

docs/ui/styling.md
```diff
@@ -165,6 +166,56 @@ Similarly to HTML, CSS can be defined inline for a UI view in the XML markup:
...
+
+{% nativescript %}
+With this pattern, a page has 3 separate stylesheets: Common, iOS and Android. For example, for page `myPage.xml` you would have 3 stylesheets:
```
tjvantoll: Common probably shouldn’t be capitalized. This is very important.

tjvantoll (commented):
I left a few minor comments, but overall so awesome that this will be actually documented now. I can’t believe this wasn’t in there already.

Also, PR #1000 🎉🍾🎊

docs/ui/styling.md
toddanglin: Good idea. Added two Playground links: one for Core and one for Angular. That said, the Playground Core example doesn't seem to be working as expected. Styles in the shared "Common" CSS file are not being imported. Any idea what I'm missing? https://play.nativescript.org/?template=play-tsc&id=w9Lt96

docs/ui/styling.md
toddanglin: This is the only change I did not address yet. The examples were short so they would fit in the numbered list. Are you suggesting that these larger examples get added in-line within the numbered list?

docs/ui/styling.md
tjvantoll: It’s the path. I hate paths.

I was able to get it working using `@import "./home/home-page.common.css";`, but I’m not sure whether that’s the best approach. `@import "../home/home-page.common.css";` also works but also feels a bit ugh.

I’ll ping the modules team to get us a good answer.

In the meantime a suggestion—you can delete files in the Playground now, and you might want to just kill the view model to help users focus on the CSS.

tjvantoll: So... I got more of a complex answer than I was hoping for with paths 😄 Seems like something might have changed to support webpack but we need to try a few things to be sure.

For now I’d probably just go with `@import "./home/home-page.common.css";`.

tjvantoll:
Merged with aef661dee81dd9fc725d2909f1e3262d3d2a0c7e and 9e342026813972ec29fb0ad2976b4e49b789b121. Thanks @toddanglin!

PR #1003, opened 2017-12-12
update of nativescript ui (rich components) page

rdlauer:
- updated all descriptions, cleaned up grammar, and clarified some points
- all new images (compressed w/ tinypng as well)


etabakov:
Looks great! My only concern is that renaming the `RadListView` to `ListView` might confuse users with the other UI Component we have in the Core Modules with the same name `ListView`. What do you think?

rdlauer:
Yep, I thought of that as well. I think it's confusing to label it as "RadListView", which is why I changed it to ListView and added a comment/note about the default {N} ListView component. Maybe that comment should be moved to be just underneath the ListView title?

docs/ui/rich-components.md
```diff
@@ -1,133 +1,130 @@
...
 ---
 
-# Rich UI Widgets
+# Professional UI Components
 
 
 NativeScript UI is a free suite of rich and customizable premium UI widgets for building NativeScript apps. It is available for download as a plugin on npmjs.com under [nativescript-pro-ui](https://www.npmjs.com/package/nativescript-pro-ui).
 
-> Continue reading more about each component below or take them up for a spin with the {% nativescript %}[NativeScript UI sample app on GitHub](https://github.com/telerik/nativescript-ui-samples){% endnativescript %}{% angular %}[NativeScript UI sample app on GitHub](https://github.com/telerik/nativescript-ui-samples-angular){%  endangular %}.
+> Continue reading about each component below or take them for a spin with the {% nativescript %}[NativeScript UI sample app on GitHub](https://github.com/telerik/nativescript-ui-samples){% endnativescript %}{% angular %}[NativeScript UI sample app on GitHub](https://github.com/telerik/nativescript-ui-samples-angular){%  endangular %}.
```
tjvantoll, line 13: Idea: we should totally get that example app working in Playground. Probably out of the scope of this PR, but it would be so cool to just link to that.

docs/ui/rich-components.md
```diff
@@ -1,133 +1,130 @@
...
-* The chart series your application needs:
+The Chart component (known as RadChart in code) can be used to visualize data in a human-readable way through lines, areas, bars, pies, and more. Some features include:
 
-	* Show trends with `Line`, `Area` and `Spline Area` charts 
-	* Compare sets of data with `Bar` charts 
-	* Illustrate proportions with `Pie` and `Donut` charts and use `Spline` and `Spline Area` charts to plot data that require the use of curve fittings
-	* Show relationship among values using `Scatter` and `Bubble` series, even use `Financial` series and indicators.
-* Annotations, tooltips and trackball
-* Multiple axes support
+* Wide array of accepted data types: numerical, string or `DataTime` data depending on the chart you want to visualize;
```
tjvantoll: `DateTime`

docs/ui/rich-components.md
```diff
@@ -1,133 +1,130 @@
...
 
 
 ## DataForm
 
+[{% nativescript %}[Documentation](http://docs.telerik.com/devtools/nativescript-ui/Controls/NativeScript/DataForm/dataform-overview){% endnativescript %}{% angular %}[Documentation](http://docs.telerik.com/devtools/nativescript-ui/Controls/Angular/DataForm/dataform-overview){% endangular %}] [{% nativescript %}[Sample Code](https://github.com/telerik/nativescript-ui-samples/tree/release/sdk/app/dataform){% endnativescript %}{% angular %}[Sample Code](https://github.com/telerik/nativescript-ui-samples-angular/tree/release/sdkAngular/app/dataform){% endangular %}]
 
-The {% nativescript %}[DataForm](http://docs.telerik.com/devtools/nativescript-ui/Controls/NativeScript/DataForm/dataform-overview){% endnativescript %}{% angular %}[DataForm](http://docs.telerik.com/devtools/nativescript-ui/Controls/Angular/DataForm/dataform-overview){% endangular %}  provides an easy and versatile approach to processing data objects’ public members and building mobile forms fast. 
+The DataForm component (known as RadDataForm in code) provides an easy and versatile approach for building mobile forms based on a provided data objects’ public members. Use DataForm to:
```
tjvantoll: This is super important, but isn’t “object’s” more appropriate than “objects’” here? You point it at a single object, and that singular object has public members.

tjvantoll (approved):
Had a couple of super trivial comments. This is sooo good 😄

tjvantoll:
@etabakov @rdlauer Yeah I’m thinking something like.

> **NOTE**: The professional ListView component is different from the ListView built in to the core NativeScript modules. Use the professional ListView if you need to take advantage of the advanced functionality listed below.

docs/ui/rich-components.md
rdlauer: Fixed!

docs/ui/rich-components.md
rdlauer: Fixed as well

rdlauer:
@tjvantoll there is a note in the ListView section already:

> RadListView is not to be confused with the [default NativeScript ListView component](https://docs.nativescript.org/ui/list-view)!

But I'm open to whatever we think makes the most sense

tjvantoll:
@rdlauer So my only concern with that is it doesn’t explain what the difference is. If you add one sentence to that helps users know which to use then 👍 from me.

rdlauer:
@tjvantoll take a look at the new version - i just used your text and moved it up under the heading

tjvantoll:
Merged :)

PR #1005, opened 2017-12-13
docs(data-binding) update data binding article

manoldonev:
- Updated code snippets
- Added {N} Playground links for most of the examples


etabakov:
:+1 for the Playground examples

tjvantoll:
Merged with 38a8f7bb1e8cb2dc262d199f10c9dae0997afb28. Thanks so much @manoldonev! The Playground examples in here are super well done, and I think this page is now a model of how we should approach documentation throughout the site.

PR #1006, opened 2017-12-14
docs(navigation): update vanilla architecture and navigation article

ADjenkov:
Fix and Update of code snippets (to be more copy pastable).

### DIFF
--- a/docs/core-concepts/navigation.md
+++ b/docs/core-concepts/navigation.md
@@ -63,9 +63,9 @@ exports.onPageLoaded = onPageLoaded;
 ```
 ``` TypeScript
 // main-page.ts
-import observableModule from "data/observable";
+import { EventData } from "data/observable";
 
-export function onPageLoaded(args: observableModule.EventData) {
+export function onPageLoaded(args: EventData): void {
     console.log("Page Loaded");
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/core-concepts/navigation.md
```diff
@@ -157,25 +158,32 @@ A more dynamic way of navigating can be done by providing a function that return
...
+const factoryFunc = (): Page => {
+    const label = new Label();
     label.text = "Hello, world!";
-    var page = new pagesModule.Page();
+    const page = new Page();
     page.content = label;
     return page;
 };
-topmost.navigate(factoryFunc);
+topmost().navigate(factoryFunc);
```
tjvantoll, line 186: Wow, good catch!

tjvantoll: Oh I see. This worked before, but the previous code was, uh, weird 😛

ADjenkov: Exactly! I personally think that this way it's more obvious how it should be used

tjvantoll:
Merged with 2a9adf5ce33830dfa886b69db6c624c46af14ecb. Thanks so much @ADjenkov! This really cleans this article up.

PR #1015, opened 2017-12-19
refactor: Update theme installation instructions

NickIliev:
The main theme imports are now handled only with **tilde** (instead of tilde-slash)
For example:
the newly introduced import syntax
```
@import '~nativescript-theme-core/css/core.dark.css';
```

the old way was
```
@import '~/nativescript-theme-core/css/core.dark.css';
```

### DIFF
--- a/docs/ui/theme.md
+++ b/docs/ui/theme.md
@@ -55,14 +55,14 @@ The NativeScript core theme includes two standard color schemes, light and dark.
 To use the light color scheme, add the following line of code to the top of your `app.css` file:
 
 ``` CSS
-@import '~/nativescript-theme-core/css/core.light.css';
+@import '~nativescript-theme-core/css/core.light.css';
 ```
 
 And in order to use the dark scheme, add this line of code to your `app.css` instead:
 
 
 ``` CSS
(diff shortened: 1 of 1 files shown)


etabakov (approved):
Looks good to me!

tjvantoll:
So I went to play with this in Playground, and I happened to noticed (because the templates do this by default), that the tilde doesn’t seem to be necessary at all. Aka this works fine with 3.3 at least.

```CSS
@import 'nativescript-theme-core/css/core.light.css';
```

Should we just drop the tilde altogether, or is that changing with 3.4?

vchimev:
@tjvantoll, @NickIliev,

starting with `nativescript-dev-webpack@0.9.0`, the tilde is necessary to resolve the theme's files into webpack builds.

This change has been reflected over the templates, for example:
https://github.com/NativeScript/template-hello-world-ng/commit/8fc4e9dc4273b35364ffd49876f1926dc40ce027#diff-a4a1075348c18f6b7dcfea46ae05fb60

vchimev (approved):

tjvantoll:
Thanks @vchimev for the clarification!

👍  from me too

PR #1020, opened 2017-12-21
Fix line break

mhrabiee:
fix line break on CSS code example.


etabakov (approved):
Thanks for the suggestion @mhrabiee!

PR #1028, opened 2017-12-29
Update action-bar.md

rajatraj733:
Website androiddrawables.com is not working. But, I could not find the alternative website. 
The website https://developer.android.com/reference/android/R.drawable.html seemed to me the one which should be followed for the system icons, but this page did not display the icons corresponding to the icon name.

### DIFF
--- a/docs/ui/action-bar.md
+++ b/docs/ui/action-bar.md
@@ -272,7 +272,7 @@ You can use the `icon` property to set an image instead of text for the action i
 
 You can use the `android.systemIcon` and `ios.systemIcon` properties to show system icons. If you define a system icon, it will be used instead of `icon` and `text` properties.
 
-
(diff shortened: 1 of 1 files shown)


etabakov:
run ci

vakrilov (changes requested):
Thanks for the PR @rajatraj733 
Too bad `androiddrawables.com` in no longer available.
Having no link with list os available values will just leave you wondering how to use the `android.systemIcon` property. Can you add in a link to ` https://developer.android.com/reference/android/R.drawable.html `. It seems good enough.

rajatraj733:
@vakrilov I have made the changes. If it requires more modification, ask me to do the same otherwise merge the pull request.

etabakov (approved):
Thanks for your effort and contribution!

PR #1032, opened 2018-01-08
docs(win-install): Update win installation

radeva:
Update win installation steps. 

(Depends on https://github.com/NativeScript/setup-scripts/pull/7 which needs to be merged first)


etabakov (approved):
Great!

PR #1039, opened 2018-01-14
Update ng-chapter-4.md

doorb02:
At this stage the property 'isLoading' is actual 'true' not 'false'.


tjvantoll:
Merged with 156ce46dc748bff6afafd7e5f80c5d9dfb8eb42f.

Thanks @doorb02!

PR #1048, opened 2018-01-19
Reference both app.routing.ts arrays

jayoma:
New Angular developers might overlook the need to inform the AppModule of this new component.


etabakov:
run ci

tjvantoll (approved):
Thanks @jasonhendee. I think this makes a lot of sense. 👍

PR #1049, opened 2018-01-22
Fixed contents link issue

surdu:


docs/publishing/publishing-ios-apps.md
```diff
@@ -19,7 +19,7 @@ previous_url: /core-concepts/releasing-apps,/core-concepts/publishing-ios-apps
 2. [Certificates, identifiers & profiles](#certificates-identifiers--profiles)
   1. [Account with Apple ID](#account-with-apple-id)
   2. [Development certificates](#development-certificates)
-  3. [Production certificates](#distribution-certificates)
+  3. [Production certificates](#distribution-provisioning-profiles)
```
etabakov: Thanks for catching this! Actually, I think the correct link should be #production-certificates, instead of #distribution-provisioning-profiles. What do you think?

surdu: I agree. Let me change that real quick

surdu:
Fixed. Now I saw I actually had copied the wrong link 😅

etabakov:
run ci

etabakov:
Thanks for the contribution 😄

surdu:
My pleasure ;)

Issue #1038, opened 2018-01-12
Resolving nsbuild01.telerek.com (ns1build01.telerek.com) .. failed name or service not found


ellipee:
i encounter this error when building on a Linux system
Resolving nsbuild01.telerek.com (ns1build01.telerek.com) .. failed name or service not found

etabakov:
Access to the build server is no longer required. Closing the issue.

Issue #954, opened 2017-11-01
Documentation should be available offline


ghost:

NickIliev:
Hey, @Fabiyo-90 the NativeScript/docs repository has detailed README.md covering all the topics from contributing to local builds using Vagrant. Let me know if you need some specific information.

[https://github.com/NativeScript/docs/blob/master/README.md](https://github.com/NativeScript/docs/blob/master/README.md)

ghost:
Why not a doc with http://devdocs.io ? It's gonna be simplier

etabakov:
Thanks for the suggestion. From what I understand DevDocs looks like a great way to browse different documentations on a single place, but it still needs an original source to scrape the content from. While we could expose our documentation through [devdocs.io](http://devdocs.io), this could not replace our current documentation: [docs.nativescript.org](http://docs.nativescript.org).

We will evaluate the enablement of our documentation through DevDocs and will eventually include it in our roadmap. Meanwhile, if you can explain what you are trying to achieve and what are you struggling with - we might be able to help you.

ghost:
I am in a country where the connection to use the internet is very expensive so I would like to have localhost documentation that is easy to use or install. I would have liked just having folder containing html files like the documentation of the framework fuelphp (www.fuelphp.com) for example, not very complicated to use. Just have the docs and click on the index.html file and everything is there.
Student from the Democratic Republic of Congo and I am a french man and my english is not good!

ghost:
I love NativeScript but I can not always connect to the internet

etabakov:
Thanks for the clarification. It looks like your requirement is to have the documentation available offline. I will rename the title of the issue to state this clearly. We will prioritise this feature and eventually include it in the roadmap. Meanwhile, you can check some tools that can make an offline copy of a website for you, like [HTTrack](https://www.httrack.com/)

ghost:
I will try to use HttTrack then report to you, I have an old version of htttrack but it hurts the copy of the site so I try to download a new version and I will tell you more

ghost:
HttTrack takes a lot of time from 8am and saw the bad connection we have in Congo, I feel that I will need a whole day is 24h!
It's very very very bad !

etabakov:
The local setup instructions are updated. Now you can setup the whole documentation locally and use it as needed.

Issue #945, opened 2017-10-20
Add IIS server to the Vagrant setup


etabakov:
Currently, when using Vagrant locally the documentation is served using NGINX server. This makes it impossible to test the IIS configuration that is part of the documentation - like redirects for example.

We should add another windows machine in the Vagrant configuration, which will be used for serving the documentation produced by the build scripts.

etabakov:
This is no longer required, as we switch the hosting servers to NGINX.

Issue #463, opened 2016-07-27
Make the documentation buildable by the community


ErjanGavalji:
There is an internal repo used to build the documentation. Rework the code so that it gets buildable and modifiable by the community.

etabakov:
Related to: https://github.com/NativeScript/docs/issues/920

etabakov:
The local instructions are updated with new setup for local usage. Moreover, it's the same as the setup user on the production environment.

PR #1060, opened 2018-02-02
Docker image for local builds

darind:
- Added docker image for local builds
- Added production nginx docker image


etabakov:
Fixes #1038 #954 #945 #920 #463

PR #1065, opened 2018-02-09
Making the documentation more applicable to its main audience; {N}ewbs!

paulpv:


docs/plugins/building-plugins.md
```diff
@@ -24,10 +24,14 @@ At their basic level NativeScript plugins are simple JavaScript modules that use
 
 ```
 nativescript-hello-world/
```
NathanaelA, line 26: I would probably change this to be something like `nativescript-coolPlugin` or something like that to signify a plugin not the stock hello world template.  :grinning:

docs/plugins/building-plugins.md
```diff
@@ -24,10 +24,14 @@ At their basic level NativeScript plugins are simple JavaScript modules that use
...
-├── index.js
-└── package.json
+└── src
+    ├── index.js
+    └── package.json
 ```
```
NathanaelA, line 30: I would probably also mention a Demo folder where the demo of the plugin resides.   
Standardized naming is "Demo" <- PAN, "Demo-ng" <- Angular demo, and "Demo-Vue" <- VueJS demo.

NathanaelA: This might be a moot point, later in the docs a demo folder is partially discussed.   :)

etabakov:
run ci

docs/plugins/building-plugins.md
radeva: Actually this structure presents the simplest plugin as it is published in npm and is correct. The `src` folder is something used during development only and described later in the article. Probably a sentence saying that this is a structure of the npm package will add more clarity in this case. What do you think?

radeva (approved):
Thanks @paulpv for the PR!

radeva:
Congratulations on this quality PR! Have a badge!
contrib

paulpv:
@radeva I just saw your comment, and then noticed it got merged anyway.
My below comments may be moot, but I thought I would respond anyway.

I wrote this PR as a stickler for starting people off using the best practices, and preventing them from starting off using bad practices that may soon shoot them in the foot.

Your comment seems catch22/chicken-egg contradictory in my mind:

> Actually this structure presents the simplest plugin as it is published in npm and is correct. The src folder is something used during development only and described later in the article. Probably a sentence saying that this is a structure of the npm package will add more clarity in this case. What do you think?

You say `The src folder is something used during development only...`, but then say that `the simplest plugin as it is published in npm ... is correct.`
Wasn't `the simplest plugin as it is published in npm` `developed` by someone?
If `the simplest plugin as it is published in npm` doesn't place its source code in the `src` subfolder then it wasn't developed using best practices and it should be updated to move its source code to a `src` folder.

{N} should not set a precedent of showing bad practices, even if just a `simplest` hello world demo, and if a bad practice is ever noticed then it should be corrected...thus the point of this PR request.

Does that sound sensible?

PR #1071, opened 2018-02-13
docs: add build script for cli docs

etabakov:


etabakov:
run ci

build/build-docs.sh
```diff
@@ -31,6 +33,9 @@ fi
 
 bundle config build.nokogiri --use-system-libraries
 
+cd $CLI_ROOT
+./docs/build-jekyll-md.sh
```
darind, line 37: This `build-jekyll-md.sh` doesn't have LF line endings specified in its `.gitattributes` file and as a result it will not work on Windows. When checked out, git will automatically convert it to CRLF line endings which break in docker.

build/docs-watcher/paths.js
```diff
@@ -11,6 +11,7 @@ module.exports.modulesRoot = path.join(module.exports.root, "..", "..", "NativeS
 module.exports.nativescriptAngularRoot = path.join(module.exports.root, "..", "..", "nativescript-angular");
 module.exports.sdkExamplesRoot = path.join(module.exports.root, "..", "..", "nativescript-sdk-examples-ng");
 module.exports.sidekickRoot = path.join(module.exports.root, "..", "..", "sidekick-docs");
+module.exports.cliRoot = path.join(module.exports.root, "..", "nativescript-cli");
```
darind: The correct path should be one more level up: `module.exports.cliRoot = path.join(module.exports.root, "..", "..", "nativescript-cli");`

build/docs-watcher/watchers.js
```diff
@@ -23,6 +23,10 @@ module.exports.WatcherService = class WatcherService {
 			basePath: path.normalize(paths.sdkExamplesRoot),
 			distPaths: [`${paths.sdkExamplesRoot}/dist/./code-samples`],
 			buildScript: `${paths.sdkExamplesRoot}/build-docs.sh`
+		}, {
+			basePath: path.normalize(paths.cliRoot),
+			distPaths: [`${paths.cliRoot}/docs-cli`],
```
darind: The dist path should be ``[`${paths.cliRoot}/./docs-cli`]``

build/docs-watcher/watchers.js
```diff
@@ -23,6 +23,10 @@ module.exports.WatcherService = class WatcherService {
 			basePath: path.normalize(paths.sdkExamplesRoot),
 			distPaths: [`${paths.sdkExamplesRoot}/dist/./code-samples`],
 			buildScript: `${paths.sdkExamplesRoot}/build-docs.sh`
+		}, {
+			basePath: path.normalize(paths.cliRoot),
```
darind: It would be better to limit the base path to ``path.normalize(`${paths.cliRoot}/docs`)`` to avoid recursive builds as the `docs-cli` content changes by the build script

etabakov:
Thanks for feedback! I think everything is addressed now.

darind:
👍 All looks good now!

PR #1084, opened 2018-02-22
Clarify open workspace for CocoaPods message

chrants:
The message was very confusing for me, as I very much so got the two mixed up still after reading this a couple of times. The wording caused me to fall into the pitfall, rather than avoid it haha. This is just to clarify to avoid it.


etabakov:
run ci

etabakov (approved):
Thanks for the contribution!

Issue #905, opened 2017-09-15
Long-form two-way data binding example doesn't work


airandfingers:
The [Two-way Data Binding](https://github.com/NativeScript/docs/blob/master/docs//tutorial/ng-chapter-3.md#----exercise-two-way-data-binding-with-angular) section of the Angular Tutorial, Chapter 3 contains this sentence:

> In the case of this example, [(ngModel)]="email" is shorthand for [text]="email" (ngModelChange)="email=$event", which binds the email element’s text attribute to an email property and adds a change event handler that updates the email property’s value whenever the user makes a change.

I tried replacing `[(ngModel)]="email"` with `[text]="email" (ngModelChange)="email=$event"`, but  `ngModelChange` didn't seem to do anything; the field still exhibited one-way data-binding.

I'm guessing this is a docs issue, not a bug.

vakrilov:
I think the correct way to do is `[text]="email" (textChange)="email=$event"`. Might be really a error in the docs.

etabakov:
Indeed, this looks like a problem in the documentation. The correct syntax should be `[text]=“email” (textChange)=“email=$event.object.text”`.

PR #1089, opened 2018-02-23
docs: fix a note for what [(ngModel)] stands for

etabakov:
https://github.com/NativeScript/docs/issues/905

### DIFF
--- a/docs/tutorial/ng-chapter-3.md
+++ b/docs/tutorial/ng-chapter-3.md
@@ -137,7 +137,7 @@ export class AppModule {}
 
 <div class="exercise-end"></div>
 
-At first glance the `[(ngModel)]` syntax looks more than a little odd, as it’s essentially a combination of the event and attribute binding syntax that you used in earlier examples. In the case of this example, `[(ngModel)]="email"` is shorthand for `[text]="email" (ngModelChange)="email=$event"`, which binds the email element’s `text` attribute to an `email` property and adds a `change` event handler that updates the `email` property’s value whenever the user makes a change.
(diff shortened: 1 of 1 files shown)


etabakov:
Verified in this Playground: https://play.nativescript.org/?template=play-ng&id=RZySOR&v=5

vakrilov (changes requested):
Can you change the quotes from `“` to `"` to avoid problems with copy-pasting

vakrilov (approved):

etabakov:
Done, thanks!

hackerunet:
I'm sorry to take my hands on this but, this seems to not been considering that commonly models are not described by just "email", a serious application would have models and references to those models exactly as angular does, I do understand that there could be some differences but, as far as I've been testing neither the documentation nor [text]="user.Email" (textChange)="user.Email=$event.object.text" works!! I got a context error in the chrome dev tools console and unfortunately I cannot trace this error far beyond. two way databinding is not easily possible in NS, if it is please consider adding some example of how to do it, at the very beginning with the simplest login page we need two way data binding and for new devs trying to get our hands dirty, is a serious fault to not have a correct documentation or real world examples how to acomplish the basic objective as update a model. Please can you be more specific how to implement 2 way DBiding using a model? thanks

vakrilov:
Hey @hackerunet - thanks for the feedback.
The `[(ngmode]]="user.Email"` should work as expected (check out this [demo in playground](https://play.nativescript.org/?template=play-ng&id=RZySOR&v=6)). If it doesn't it is probably a bug - you can open an issue about it. Note: that it will not work is the `email` is null or undefined, so make sure you initialize this in components constructor or `ngOninit`. 
About the docs - [later on in the chapter](https://docs.nativescript.org/angular/tutorial/ng-chapter-3#33-structuring-your-app) we switch to using a `user` object which might be the thing you are looking for.

Issue #856, opened 2017-08-02
tutorial/ng-chapter-6.md says NativeScript doesn't expose placeholder-color, but it does now


madakk:
This was finally resolved under https://github.com/NativeScript/NativeScript/issues/712 and now is supported, so the text should probably be changed to say that while this can be done with the provided CSS attribute, this is how you'd do it in the native way instead.

tjvantoll:
Thanks for reporting this @madakk! Can’t believe this hasn’t come up before actually 😄 

I’m adding the “help wanted” label in case anyone from the community wants to tackle this. Otherwise I’ll change the wording up when I get from vacation at the end of next week.

radeva:
I'm closing this issue since it is resolved with the above PR.

PR #1090, opened 2018-02-23
docs: remove statement that placeholder color cannot be set with CSS

etabakov:
Mentioned in https://github.com/NativeScript/docs/issues/856


tjvantoll:
Looks good @etabakov. Thanks!

PR #1095, opened 2018-03-10
Workaround for Javascript heap out of memory errors

noumaans:
Added workaround to deal with heap out of memory issues, which will happen for a reasonable sized app built on {N}.
Please see https://github.com/NativeScript/nativescript-dev-webpack/issues/458

**NOTE** Windows environment variable setting has not been tested by me - please confirm / verify it for Windows. macOS / Linux has been verified to work.

### DIFF
--- a/docs/best-practices/bundling-with-webpack.md
+++ b/docs/best-practices/bundling-with-webpack.md
@@ -417,6 +417,19 @@ const myPlugin = "my-plugin";
 global.loadModule(myPlugin);
 ```
 
+### Debugging Bundling Errors
+
+#### Javascript heap out of memory
+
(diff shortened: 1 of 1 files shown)


etabakov:
run ci

sis0k0 (approved):

etabakov (commented):
Thanks for the addition!

JohnGalt1717:
This is great in the documentation, however this should be set by default by tns and it should be 8192.

NickIliev:
@JohnGalt1717 setting larger value by default could be problematic on PCs with not enough memory

JohnGalt1717:
@NickIliev worst case is that it just runs out of memory on the machine. But as it is now, you can't build no matter how much memory you have without hacks which cause problems on shared build hosts.

PR #1096, opened 2018-03-16
Updated native iOS component for HtmlView

ikoevska:
See https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/html-view/html-view.d.ts#L22 and https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/ui/html-view/html-view.ios.ts#L12 Widget actually uses UITextView and not UILabel


etabakov:
run ci

etabakov (approved):
Thanks for the update, Iva! Nice to see you around :)

ikoevska:
My pleasure :)

PR #1103, opened 2018-03-24
Add instructions for installing AVD Manager

chrismullins:
This is a common issue in Android Studio versions >3.0.0, and is not yet added to the official documentation.

### DIFF
--- a/docs/tooling/android-virtual-devices.md
+++ b/docs/tooling/android-virtual-devices.md
@@ -9,6 +9,8 @@ slug: android-emulators
 
 Follow the official documentation on [Creating and Managing Virtual Devices](https://developer.android.com/studio/run/managing-avds.html), where the process of downloading, setting up, and using Android Emulators via Android studio is covered.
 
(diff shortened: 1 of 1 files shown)


chrismullins:
I also found that the installation process in the [Nativescript Quick Setup section for Mac OS](https://docs.nativescript.org/start/quick-setup#macos) uses brew to install the Android SDK, whereas if the user chooses to install Android Studio (which includes the AVD manager needed in this step) it will install the Android SDK into a separate directory (`~/Library/Android/sdk`). This will cause the `PANIC: Cannot find AVD system path. Please define ANDROID_SDK_ROOT` error when the user tries to launch an android device from the AVD Manager (despite the correct environment variables being set).

Uninstalling via brew as this answer prescribes in [1] fixes this problem.

There may be a better way to structure this, I'm just trying to document this feedback as I come across problems so I don't forget :)

[1] https://stackoverflow.com/questions/39645178/panic-broken-avd-system-path-check-your-android-sdk-root-value#answer-45490757

etabakov:
run ci

etabakov:
run ci

etabakov (approved):

PR #1104, opened 2018-03-27
Add information about proxy

tadeucariolano:
Add information about using sdkmanager behind a corporate proxy


etabakov:
run ci

etabakov:
That's a good suggestion. Can you add reference to the SDK manager documentation, instead of directly explaining the options? This would be easier for us for maintenance. Thanks for the effort!

tadeucariolano:
Sure. I'll do it.

PR #1107, opened 2018-03-30
Fixed small code typo

surdu:


NickIliev:
@surdu - why do you think this is a typo?

The callback methods can have a custom name and is better to have a descriptive custom name (e.g. like `onPageLoaded`, `onLabelLoaded`, `onStackUnloaded`) so we could have a better context where the callback have originated from. Of course, the event should be referenced as named by NativeScript(e.g. `loaded`, `unloaded`, `navigatedTo`, etc.)

surdu:
@NickIliev Totally agree, but I believe it's a typo because at the bottom of the code we export the function as `exports.onLoaded = onLoaded;`.

NickIliev:
Oh my.. nice catch! :)

NickIliev (approved):

etabakov:
run ci

PR #1108, opened 2018-03-30
Navigation Transitions: fix object context for custom transitions

nicoeg:
The use of arrow functions for the `createAndroidAnimator ` and `animateIOSTransition ` callbacks prevented binding of the scope which is needed for `getDuration` and `getCurve` functions.
I have changed the callbacks to use normal functions instead in the examples to avoid this.


NickIliev (approved):

etabakov:
run ci

etabakov (approved):
Thanks for the effort!

PR #1109, opened 2018-04-03
Update publishing-android-apps.md

marablayev:
Apk location in platfroms file


etabakov (approved):
Good catch! Thanks for the contribution.

Issue #995, opened 2017-12-07
Feedback form


tsonevn:
- [x]  NS + NS Angular example - how to use `toBase64String` method - https://docs.nativescript.org/cookbook/image-source - done via [https://github.com/NativeScript/NativeScript/pull/5390](https://github.com/NativeScript/NativeScript/pull/5390) and via [https://github.com/NativeScript/nativescript-sdk-examples-ng/pull/265](https://github.com/NativeScript/nativescript-sdk-examples-ng/pull/265)

- [x] extended example of how to change the Activity Indicator state on receiving data from HTTP request (NS + NS Angular example) **(in progress)**

- [x]  providing a description how the [imageSource module](https://docs.nativescript.org/cookbook/image-source) and it's methods works.

- [x] creating sample non-Angular app, which provides some use case examples- similar to [nativescript-sdk-examples-ng](https://github.com/NativeScript/nativescript-sdk-examples-ng/tree/master/app)

- [x] [example ](https://docs.nativescript.org/cookbook/ui/frame) how to use the `BackstackEntry` option.

- [x] Improve the **ListView** example snippets in the documentation - related to [https://github.com/NativeScript/docs/issues/981](https://github.com/NativeScript/docs/issues/981)

- [x] Improve the **TabView** example snippets in the documentation - https://docs.nativescript.org/cookbook/ui/tab-view

- [x] Improve the **TextField** example for N + Angular (sdk examples) - related to [https://github.com/NativeScript/docs/issues/987](https://github.com/NativeScript/docs/issues/987)
Resolved via [https://github.com/NativeScript/nativescript-sdk-examples-ng/pull/260](https://github.com/NativeScript/nativescript-sdk-examples-ng/pull/260)

- [x] improve **Switch** examples [https://docs.nativescript.org/cookbook/ui/switch](https://docs.nativescript.org/cookbook/ui/switch)

- [x] improve the **GridLayout** examples [https://docs.nativescript.org/cookbook/ui/layouts/grid-layout](https://docs.nativescript.org/cookbook/ui/layouts/grid-layout)

- [x] improve the **WrapLayout** examples [https://docs.nativescript.org/cookbook/ui/layouts/wrap-layout](https://docs.nativescript.org/cookbook/ui/layouts/wrap-layout)

- [ ] Research - what are the available options for providing live demo inside the documentation

- [x] providing sample project for the [Multithreading Model article](https://docs.nativescript.org/core-concepts/multithreading-model)
[...]

PR #1114, opened 2018-04-04
docs: Android image optimization article update (decode properties)

NickIliev:
- Android image optimization article update
- Include `decodeWidth` and `decodeHeight` detailed information
- Article "Create Android Emulators" revision related to feedback from requests [https://github.com/NativeScript/docs/issues/995#issuecomment-358571662](https://github.com/NativeScript/docs/issues/995#issuecomment-358571662)


docs/tooling/android-virtual-devices.md
```diff
@@ -5,754 +5,70 @@ position: 60
...
-
-
-
-
-
-
+# Setup Android Emulators
 
+Apart from using real Android devices, a viable option is to download, install and use an Android emulator.
+In NativeScript, we can use all Android emulators that are connected and recognized by `tns devices` command.
```
etabakov: We can refer here to the [`tns device` documentation article](https://docs.nativescript.org/docs-cli/device/device).

etabakov (commented):
Looks good to me!

docs/best-practices/images-optimisations.md
```diff
@@ -8,20 +8,45 @@ slug: images-performance
 # Android Image Optimization
 
 One of the most common scenarios for modern mobile applications is to work with multiple images often in high definition formats.
-It is crucial for each mobile developer to handle memory related issues and optimize the application so it could handle large data (e.g. API call which will load hundreds of photos). 
+It is crucial for each mobile developer to handle memory related issues and optimize the application so it could handle large data (e.g., API call which will load hundreds of photos).
```
pedromorgan: It is very important for mobile developer to handle memory related issues and optimize an application so it can handle large data(e.g. a web request to that downloads hundreds of photos and alike).

pedromorgan (commented):
I changed "crucial" to very important... In japanese it would have other context... and other langs

MartoYankov:
I think the sequence of the information about image optimization should be changed a bit, so that the message is more clear. The bullet point about what happens when decodes are not set should be moved to `Handling large images and avoiding Out Of Memory exception` since this is now the default behavior for Android.

I think our message should be - We made optimizations, so normally you won't need to set decodes now. However, if you still experience OOM or want a fine grain control you can set the decodes. And the info about the new decodes should be afterward.

docs/best-practices/images-optimisations.md
NickIliev: Thanks for the comment - updating the PR immediately.

docs/tooling/android-virtual-devices.md
```diff
@@ -5,754 +5,70 @@ position: 60
...
-
-
-
-
-
-
+# Setup Android Emulators
 
+Apart from using real Android devices, a viable option is to download, install and use an Android emulator.
+In NativeScript, we can use all Android emulators that are connected and recognized by `tns device` command. Detailed information about the `device` command can be found [here]((https://docs.nativescript.org/docs-cli/device/device)).
```
etabakov: According to the [best practices](https://usabilitygeek.com/hyperlink-usability-guidelines-usable-links/), the text of the hyperlink should explain what the hyperlink is about. With that said 'here' is probably not a good option. Think about something like "More information about this command could be found in the [`tns device` article](https://docs.nativescript.org/docs-cli/device/device)"

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
 # Android Image Optimization
 
 One of the most common scenarios for modern mobile applications is to work with multiple images often in high definition formats.
-It is crucial for each mobile developer to handle memory related issues and optimize the application so it could handle large data (e.g. API call which will load hundreds of photos). 
+It is essential for the mobile developer to handle memory related issues and optimize an application so it can process large data(e.g., a web request to that downloads hundreds of photos and alike).
```
ggarabedian: Remove **to** -> (...) e.g., a web request **to** that downloads (...)

I personally prefer using **for example,** instead of **e.g.,** because a lot of people are not familiar with the abbreviation. Also, you can consider changing **and alike** to **and other similar operations** or remove it altogether as you have already mentioned that this is just an example.

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
 
-In this article, we will take a look at how Image module works in NativeScript and 
+In this article, we will take a look at how the `Image` module works in NativeScript and 
 cover the techniques that will improve Android application performance.
 
+* [Handling large images](#handling-large-images-and-avoiding-out-of-memory-exception)
```
ggarabedian: I recommend you use the entire title of the article or section when linking it:
[Handling large images and avoiding Out Of Memory exception]

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
+* [Using `loadMode` property](#using-loadmode-property)
+* [Using `useCache` property](#using-usecache-property)
+
 ## Handling large images and avoiding Out Of Memory exception
 
-In some cases when working with multiple large images on devices with low memory, an `Out Of Memory` exception can occur. To prevent that scenario, in NativeScript 2.5.x and above using the `src` property in Android will internally load the Bitmap in Java. Bitmap memory stays in Java world and reclaims once the Bitmap is no longer in use (e.g. there is no need for the Javascript object to be collected).This way Bitmap memory management is not an issue.
+In some cases when working with multiple large images on devices with low memory, an `Out Of Memory` exception can occur. To prevent that scenario, in NativeScript 2.5.x and above using the `src` property in Android will internally load the Bitmap in Java. Bitmap memory stays in Java world and reclaims once the Bitmap is no longer in use (e.g., there is no need for the Javascript object to be collected).This way Bitmap memory management is not an issue.
```
ggarabedian: Change **reclaims** to **it is reclaimed** (the **it** is optional) -> (...) Java world and reclaims once the Bitmap is no longer in use (...). 

Consider using **recover** (is recovered) or **restore** (is restored) instead of **reclaim**.

A missing space between a dot and the next sentence -> (...) collected).This way Bitmap (...)

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
 
 In contrast, when using `ImageSource` or Base64 encoded string, the Bitmap is transferred to Javascript, so it will be released when Javascript object reclaims. Javascript garbage collection happens less frequently than Java garbage collection which might lead to Out Of Memory.
 
-> **Tip**: Use `src` property of your `Image` to set your images to avoid Out Of Memory related issues.
+> **Recommendation**: Use `src` property of your `Image` to set your images to avoid `Out Of Memory` related issues.
```
ggarabedian: Add **the** before src -> (...) Use `src` property (...).

You may consider restructure the entire sentence so it first states what problem will be solved and then how to solve it. Looks a bit more clear to me, but this is personal preference. For example:
To avoid `Out of Memory` related issues, use the `src` property of your `Image` to set your images.

Why have you decided to change **Tip** with **Recommendation**? I am asking because **Tip** is commonly used in such cases. If we have decided to move to **Recommendation**, please let me know, so I can start using it in the Sidekick docs as well.

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
+
+### Using `decodeHeight` and `decodeWidth` properties
+
+As an additional feature for Android, NativeScript supports `decodeWidth` and `decodeHeight`. These properties will **downsample** your image so that it will take less memory. 
+The goal is to avoid as much as possible out of memory exceptions caused by images being loaded into memory and at the same time display crispy images. 
+
+> **Recommendation:** Use `decodeWidth` and `decodeHeight` only when working with large photos are there are `Out of Memory` exceptions issues. With NativeScript 3.x.x and above, image optimizations were implemented and in the common scenarios, you should not worry about hitting OOM.
```
ggarabedian: I am not sure about this one, but I believe that the first **are** needs to be an **and** -> (...) when working with large photos **are** there are `Out of Memory` (...).

You should also consider removing **exceptions** or **issues** from the same sentence. I would recommend keeping only **issues**.

If you plan to abbreviate a word (the **OOM** in the last sentence), I would recommend mentioning the abbreviation in parenthesis after the full name somewhere in the text. I know that OOM is used widely, nevertheless, in my opinion, it's good to have it. For example:
(...) are there are `Out of Memory` **(OOM)** issues (...)

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
+### Using `decodeHeight` and `decodeWidth` properties
+
+As an additional feature for Android, NativeScript supports `decodeWidth` and `decodeHeight`. These properties will **downsample** your image so that it will take less memory. 
+The goal is to avoid as much as possible out of memory exceptions caused by images being loaded into memory and at the same time display crispy images. 
+
+> **Recommendation:** Use `decodeWidth` and `decodeHeight` only when working with large photos are there are `Out of Memory` exceptions issues. With NativeScript 3.x.x and above, image optimizations were implemented and in the common scenarios, you should not worry about hitting OOM. 
+
+When working with the decode properties, the following considerations should be taken:
+
+- The `decodeWidth` and `decodeHeight` properties accepts **DIP** (device independent pixels) as measurement units. This means that image decoding will happen concerning device DPI.
```
ggarabedian: accept -> (...) `decodeHeight` properties **accepts** **DIP** (...)

If I understand the last sentence correctly, you should consider changing **concerning** with **based on the**.

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
 
-> **Tip**: When working with large images, use `decodeWidth` and `decodeHeight` to downsample the image. Use `loadMode="async"` to prevent blocking of the UI while the image is loading.
+- Image caching now takes into account the `decodeWidth` and `decodeHeight` values. Identical images with different decode property values will now be retrieved and saved separately in the cache. This results in better quality images - if you have a small version of the image in a master list and want to decode it with 100 x 100 DP, and then want to display it in 1000 x 1000 DP on the detail page, the detailed image will now not be blurry. This also means you can now control caching - using the same image with the same decode parameter values will still get the image from the cache.
+
+> **Important**: The properties `decodeWidth`, `decodeHeight`  will work only for Android. Setting them for our iOS images will not change the application behaviour in any way.
```
ggarabedian: (...) The `decodeWidth` and `decodeHeight` properties will work only for Android. (...)

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
+
+### Using `loadMode` property
+
+With [loadMode](http://docs.nativescript.org/api-reference/modules/_ui_image_.html#loadmode) set to `async`, the image will load asynchronously meaning the UI won't block by the decoding and preloading operations. The developers can use `loadMode` on both iOS and Android.
```
ggarabedian: Consider changing **meaning** to **, which means**. For example:
(...) will load asynchronously, which means (that) the UI won't be blocked by the (...)

docs/tooling/android-virtual-devices.md
```diff
@@ -5,754 +5,70 @@ position: 60
...
+```
 
+> **Tip:** Sometimes emulators take longer to start. As a recommendation and to avoid timing issues, start the emulator before executing other CLI commands.
+Once the emulator is started, leave it open to avoid the initial load time the next time you need to deploy an Android application.
 
+## Creating Android Virtual Device via Android Studio
 
+Follow the official documentation on [Creating and Managing Virtual Devices](https://developer.android.com/studio/run/managing-avds.html), where the process of downloading, setting up, and using Android Emulators via Android studio is covered.
 
+> **Note:** if your top-level "Tools" menu does not include the "Android" option (a common issue in versions >3.0.0), follow the instructions in [this StackOverflow answer](https://stackoverflow.com/questions/46948322/how-to-open-avd-manager-in-android-studio-3-0-version) to add this to the toolbar and enable the AVD Manager.
```
ggarabedian: Capitalize the **if**

docs/tooling/android-virtual-devices.md
```diff
@@ -5,754 +5,70 @@ position: 60
...
 
 
+## Creating Android Virtual Device via command line tool
 
+The `avdmanager` is a tool that allows you to create and manage Android Virtual Devices (AVDs) from the command line.
+The `avdmanager` is provided in the Android SDK Tools package (25.3.0 and higher) and is located in `<ANDROID_HOME_PATH_HERE>/tools/bin/`.
+Follow the official documentation on [creating AVD via avdmanager and command line](https://developer.android.com/studio/command-line/avdmanager.html).
```
ggarabedian: It looks like each sentence is on a new line. Is this intentional? It would be better if all 3 sentences are in one paragraph.

Also, consider rephrasing the last sentence. For example:
For more information about the `avdmanager` and how to use it to create AVDs, see the [official avdmanager documentation](https://developer.android.com/studio/command-line/avdmanager.html).

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
 # Android Image Optimization
 
 One of the most common scenarios for modern mobile applications is to work with multiple images often in high definition formats.
-It is crucial for each mobile developer to handle memory related issues and optimize the application so it could handle large data (e.g. API call which will load hundreds of photos). 
+It is essential for the mobile developer to handle memory related issues and optimize an application so it can process large data(for example, a web request that downloads hundreds of photos and alike).
```
etabakov: Missing space between 'data' and '('

docs/best-practices/images-optimisations.md
```diff
@@ -8,22 +8,49 @@ slug: images-performance
...
 
-> **Tip**: When working with large images, use `decodeWidth` and `decodeHeight` to downsample the image. Use `loadMode="async"` to prevent blocking of the UI while the image is loading.
+- Image caching now takes into account the `decodeWidth` and `decodeHeight` values. Identical images with different decode property values will now be retrieved and saved separately in the cache. This results in better quality images - if you have a small version of the image in a master list and want to decode it with 100 x 100 DP, and then want to display it in 1000 x 1000 DP on the detail page, the detailed image will now not be blurry. This also means you can now control caching - using the same image with the same decode parameter values will still get the image from the cache.
+
+> **Important**: The `decodeWidth` and `decodeHeight` properties will work only for Android.. Setting them for our iOS images will not change the application behaviour in any way.
```
etabakov: Redundant second dot?

etabakov (approved):

ggarabedian (approved):

PR #1118, opened 2018-04-06
docs: add application.run to application lifecycle article

NickIliev:
Article and snippet revised with `application.run`

### DIFF
--- a/docs/core-concepts/application-lifecycle.md
+++ b/docs/core-concepts/application-lifecycle.md
@@ -36,7 +36,7 @@ import { Component } from '@angular/core';
 import { AppComponent } from './app.component';
 ```
 
-Some modules are libraries of other modules. Modules installed as npm packages (like `@angular/core` in the above example) should be referenced without a path prefix. When we import from one of our own files, we prefix the module name with the file path. In this example we specify a relative file path (./). That means the source module is in the same folder (./) as the module importing it. 
+Some modules are libraries of other modules. Modules installed as npm packages (like `@angular/core` in the above example) should be referenced without a path prefix. When we import from one of our own files, we prefix the module name with the file path. In this example, we specify a relative file path (./). That means the source module is in the same folder (./) as the module importing it. 
 
 ## Components
 
@@ -46,7 +46,7 @@ Components are the fundamental building blocks of NativeScript applications buil
 * A component knows how to render itself.
 * A component configures dependency injection.
 * A component has a well-defined public API of input and output properties.
-* A component has well-defined lifecycle.
+* A component has a well-defined lifecycle.
 
 ### Component example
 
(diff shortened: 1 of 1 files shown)


docs/core-concepts/application-lifecycle.md
```diff
@@ -36,7 +36,7 @@ import { Component } from '@angular/core';
 import { AppComponent } from './app.component';
 ```
 
-Some modules are libraries of other modules. Modules installed as npm packages (like `@angular/core` in the above example) should be referenced without a path prefix. When we import from one of our own files, we prefix the module name with the file path. In this example we specify a relative file path (./). That means the source module is in the same folder (./) as the module importing it. 
+Some modules are libraries of other modules. Modules installed as npm packages (like `@angular/core` in the above example) should be referenced without a path prefix. When we import from one of our own files, we prefix the module name with the file path. In this example, we specify a relative file path (./). That means the source module is in the same folder (./) as the module importing it.
```
ggarabedian: "Some modules are libraries of other modules" sounds a bit strange. I know this is not written now, but you may want to consider changing it. For example: 
"Some (of the) modules are/can be dependencies of other modules."
"Some (of the) modules can depend on one or more separate modules."

docs/core-concepts/application-lifecycle.md
```diff
@@ -139,17 +139,20 @@ This method is required only for iOS applications.
...
 iOS calls UIApplication and triggers the application main event loop.
 */
 
-import { start as applicationStart } from "application";
-applicationStart({ moduleName: "main-page" });
+import * as application from "tns-core-modules/application";
+application.run({ moduleName: "main-page" });
 ```
+
+> **Note:** Prior to version 4.0.0 all NativeScript application had single topmost `Frame` implicitly created by the `application.start()` method. With NativeScript 4.x.x and above, you can create multiple `Frame` instances using the `application.run()` method. More about the `Frame` API and navigation could be found in the [applicaiton lifecycle article](https://docs.nativescript.org/core-concepts/navigation)
```
ggarabedian: application -> (...) could be found in the [**applicaiton** lifecycle article] (...)

ggarabedian (approved):
Only one small typo to fix. Looks good to me 👍

vakrilov (approved):

docs/core-concepts/application-lifecycle.md
```diff
@@ -36,7 +36,7 @@ import { Component } from '@angular/core';
 import { AppComponent } from './app.component';
 ```
 
-Some modules are libraries of other modules. Modules installed as npm packages (like `@angular/core` in the above example) should be referenced without a path prefix. When we import from one of our own files, we prefix the module name with the file path. In this example we specify a relative file path (./). That means the source module is in the same folder (./) as the module importing it. 
+Some of the modules can depend on one or more separate modules.. Modules installed as npm packages (like `@angular/core` in the above example) should be referenced without a path prefix. When we import from one of our own files, we prefix the module name with the file path. In this example, we specify a relative file path (./). That means the source module is in the same folder (./) as the module importing it.
```
MartoYankov: Two dots after first sentence.

PR #1120, opened 2018-04-06
update Page docs to be better

pedromorgan:


pedromorgan:
Dtill vonflicting with myself re homepage..

- is is homepage
- or index page
- or start page..

PR #1122, opened 2018-04-07
path dialogs and dev casting

pedromorgan:
Make is case insentive and dev friendly eg Alert != alert() or Alert() etc...


NickIliev (approved):

NickIliev (commented):
I guess we could use the same convention for all dialog options (e.g. `login` to become `login()` and `action` to become `action()`)

etabakov:
@pedromorgan Would you mind to address the feedback from @NickIliev, so we can merge this PR?

etabakov:
Closing the PR due to inactivity.

PR #1123, opened 2018-04-10
docs: revise navigation article

NickIliev:
Snippets and content revised to comply with NativeScript 4.0.0 changes in the Frame API.


docs/core-concepts/application-lifecycle.md
```diff
@@ -120,17 +120,17 @@ platformNativeScriptDynamic().bootstrapModule(AppModule).then(() => {
...
 * [Use Application Events](#use-application-events)
 * [Android Activity Events](#android-activity-events)
 * [iOS UIApplicationDelegate](#ios-uiapplicationdelegate)
 * [Persist and Restore Application Settings](#persist-and-restore-application-settings)
 
-## Start Application
+## Application Run
 
-This method is required only for iOS applications. 
+The method `run` from the `application` module is required to start the application and accepts an entry point as `NavigationEntry` argument.
```
vakrilov: It is `NavigationEntry` or path to the `app-root.xml` file

docs/core-concepts/application-lifecycle.md
```diff
@@ -139,17 +139,20 @@ This method is required only for iOS applications.
 iOS calls UIApplication and triggers the application main event loop.
 */
 
-var application = require("application");
-application.start({ moduleName: "main-page" });
+const application = require("tns-core-modules/application");
+application.run({ moduleName: "main-page" });
```
vakrilov: Use `app-root` instead of `main-page`

docs/core-concepts/application-lifecycle.md
```diff
@@ -139,17 +139,20 @@ This method is required only for iOS applications.
...
 ```
 ``` TypeScript
 /*
 iOS calls UIApplication and triggers the application main event loop.
 */
 
-import { start as applicationStart } from "application";
-applicationStart({ moduleName: "main-page" });
+import * as application from "tns-core-modules/application";
+application.run({ moduleName: "main-page" });
```
vakrilov: Use `app-root` instead of `main-page`

docs/core-concepts/application-lifecycle.md
```diff
@@ -139,17 +139,20 @@ This method is required only for iOS applications.
...
 ```
+
+> **Note:** Prior to version 4.0.0 all NativeScript application had single topmost `Frame` implicitly created by the `application.start()` method. With NativeScript 4.x.x and above, you can create multiple `Frame` instances using the `application.run()` method. More about the `Frame` API and navigation could be found in the [applicaiton lifecycle article](https://docs.nativescript.org/core-concepts/navigation)
```
vakrilov: Change the focus to the ability to specify root view. For example:
> Prior to version 4.0.0 all NativeScript application had single topmost `Frame` implicitly created by the `application.start()` method. With NativeScript 4.x.x and above the root `Frame` is no longer implicitly created. Instead you can specify any `View` to be the root of your application.  More about ...

docs/core-concepts/application-lifecycle.md
```diff
@@ -219,21 +234,19 @@ application.on(application.lowMemoryEvent, function (args) {
...
-    } else if (args.ios) {
-        // For iOS applications, args.ios is NativeScriptError.
-        console.log("NativeScriptError: " + args.ios);
-    }
+application.on(application.uncaughtErrorEvent, (args) => {
+    console.log("Error: " + args.error);
 });
 
-application.start({ moduleName: "main-page" });
+application.run({ moduleName: "main-page" });
```
vakrilov: use `app-root`

docs/core-concepts/application-lifecycle.md
```diff
@@ -423,7 +440,7 @@ if (application.android) {
     });
 }
 
-application.start();
+application.run({ moduleName: "main-page" });
```
vakrilov: use `app-root`

docs/core-concepts/application-lifecycle.md
```diff
@@ -565,7 +582,7 @@ class MyDelegate extends UIResponder implements UIApplicationDelegate {
     }
 }
 ios.delegate = MyDelegate;
-applicationStart({ moduleName: "main-page" });
+applicationRun({ moduleName: "main-page" });
```
vakrilov: use `app-root`

docs/core-concepts/navigation.md
```diff
@@ -69,98 +72,213 @@ export function onPageLoaded(args: EventData): void {
     console.log("Page Loaded");
 }
 ```
-**Create a page in code**
 
-To apply this approach, you need to create a function named `createPage` that will return an instance of your page. NativeScript considers `createPage` a factory function.
+#### Create a page in code
```
vakrilov, line 76: @NickIliev What do you think of removing this section at all? Using a module with `createPage` is not a best practice.

docs/core-concepts/navigation.md
```diff
@@ -69,98 +72,213 @@ export function onPageLoaded(args: EventData): void {
...
+</GridLayout>
+```
 
-In addition, each `Page` instance carries information about the frame object which navigated to it in the `frame` property. This lets you navigate with the `frame` property as well. 
+## Navigation
 
+The [`Frame`](http://docs.nativescript.org/api-reference/classes/_ui_frame_.frame.html) class represents the logical unit that is responsible for navigation between different pages. An application can have single or multiple `Frame` instances depending on the business logic and requirements. 
 
-### The topmost frame
+### Getting Frame reference
```
vakrilov: Change `reference` to `Reference`

docs/core-concepts/navigation.md
```diff
@@ -69,98 +72,213 @@ export function onPageLoaded(args: EventData): void {
...
+```
+``` TypeScript
+// app-root.ts
+import { getFrameById } from "tns-core-modules/ui/frame";
+const frame = getFrameById("firstFrame");
+frame.navigate("second/second-page");
+```
+```XML
+<!-- app-root.xml -->
+<Frame id="firstFrame" defaultPage="home/home-page"/>
```
vakrilov, line 271: Separate the XML example form the above JS/TS examples with something like "You can do the same declaratively in XML like this:". 

These are different ways for doing the same thing. If the code snippets are right one after the another it might give the impression that you need *both* the code and the XML.

docs/core-concepts/navigation.md
```diff
@@ -169,12 +287,14 @@ const factoryFunc = () => {
     page.content = label;
     return page;
 };
-topmost().navigate(factoryFunc);
+
+const frame = frameModule.getFrameById("firstFrame");
+frame.navigate(factoryFunc);
```
vakrilov: Can you change those examples to using navigation entry with `create` function. Like this:
```
const frame = frameModule.getFrameById("firstFrame");
frame.navigate({ 
    create: () => {
        const label = new labelModule.Label();
        label.text = "Hello, world!";
        const page = new pagesModule.Page();
        page.content = label;
        return page;
    }
});
```
I think it is more clear this way.

docs/core-concepts/navigation.md
```diff
@@ -183,52 +303,58 @@ const factoryFunc = (): Page => {
     page.content = label;
     return page;
 };
-topmost().navigate(factoryFunc);
+
+const frame = getFrameById("firstFrame");
+frame.navigate(factoryFunc);
```
vakrilov: Same as the above comment

docs/core-concepts/navigation.md
```diff
@@ -776,10 +909,10 @@ All the values in screen size qualifiers are in density independent pixels(dp) &
...
 
-The platform qualifiers are executed during build time, while the others are executed during run time. For example, the app.ios.css file will not be taken in consideration when building for the Android platform. Contrary, the screen size qualifiers will be considered just after the application runs on a device with a specific screen size. 
+The platform qualifiers are executed during build time, while the others are executed during runtime. For example, the app.ios.css file will not be taken into consideration when building for the Android platform. Contrary, the screen size qualifiers will be considered just after the application runs on a device with specific screen size. 
 
 ### Orientation qualifiers
 * `land` - orientation is in landscape mode.
 * `port` - orientation is in portrait mode.
 
-> Note: All qualifiers are taken into account when the page is loading. However, changing the device orientation will not trigger page reload and will not change the current page.
+> Note: All qualifiers are taken into account when the page is loading. However, changing the device orientation will not trigger a page reload and will not change the current age.
```
vakrilov: Change `age` -> `page` :)

vakrilov (changes requested):

docs/core-concepts/navigation.md
NickIliev: The example is in fact, showing how to navigate though frame reference from the default `home-page` to a `second-page` so the XML is part of the same example to demonstrate the `id="firstFrame"`. Hwever, I will add a simpler example demonstraing only the `defaultPage` property and using only XML `Frame` element

vakrilov (approved):

PR #1164, opened 2018-05-16
refactor: Android metadata section

NickIliev:
Removed dead link to a non-existing article about the metadata generator tool (that is now part of the Android runtime).  Related to a feedback form thread.

Added note about using `require` vs `import`


etabakov (approved):
That's great. I noticed that this article has multiple h1 headings. There can be only one, so you might consider "downgrading" the headings with one level.

tsonevn (approved):

PR #1178, opened 2018-06-01
refactor: add modal navigation example

NickIliev:
Modal page navigation example in Core section.


MartoYankov:
@NickIliev You would also need a modal-root.xml where you have a Frame with defaultPage leading to modal-page. The idea is the developer opens modal-root as a modal view.

PR #1213, opened 2018-06-22
refactor: addaptive icons note added + improved article structure

NickIliev:
Add: note about adaptive icons

Improve: removed the manual inside TOC (auto TOC generated in the new structure)


etabakov (approved):
👍

Issue #1197, opened 2018-06-20
Mention about calling variadic functions


Tyg-g:
### Make sure to check the existing issues in this repository 
Yes
### Please, tell us what's the problem?
[ ] A typo
[ ] Wrong documentation
[X] Improvement of existing article
[ ] Missing documentation
[ ] New article needed

### Please, tell us more details
Describe how to call different variadic functions in [JavaScript to Java](https://docs.nativescript.org/runtimes/android/marshalling/js-to-java) article for Android, and also for iOS if it is implemented.

NickIliev:
iOS [doesn't support variadic methods](https://docs.nativescript.org/core-concepts/ios-runtime/Limitations)

NickIliev:
Article improved for Android - Resolved via [https://github.com/NativeScript/docs/pull/1239](https://github.com/NativeScript/docs/pull/1239)

PR #1239, opened 2018-07-03
refactor: add marshalling examples

NickIliev:
Added marshaling examples for arrays of primitive types and jagged arrays

Related to [https://github.com/NativeScript/docs/issues/1197](https://github.com/NativeScript/docs/issues/1197)

### DIFF
--- a/docs/core-concepts/android-runtime/marshalling/java-to-js.md
+++ b/docs/core-concepts/android-runtime/marshalling/java-to-js.md
@@ -147,11 +147,11 @@ var objectArr = Array.create(java.lang.Object, 10);
 ```
 
 ### Array of Primitive Types
-The automatic marshalling works only for cases with arrays of objects. In cases where you have a method that takes an array of primitive types, we need to convert as follows:
+The automatic marshalling works only for cases with arrays of objects. In cases where you have a method that takes an array of primitive types, you need to convert it as follows:
 ```Java
 public static void myMethod(int[] someParam)
 ```
-Then we need to invoke it as follows:
+Then yoy need to invoke it as follows:
 ```JavaScript
 let arr = Array.create("int", 3);
 arr[0] = 1;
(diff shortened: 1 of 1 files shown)


docs/core-concepts/android-runtime/marshalling/java-to-js.md
```diff
@@ -137,15 +138,48 @@ var jaggedBooleanArray3 = Array.create("[[Z", 10);
...
 // equivalent to String[] stringArr = new String[10];
 var stringArr = Array.create(java.lang.String, 10);
 
 // equivalent to Object[] objectArr = new Object[10];
 var objectArr = Array.create(java.lang.Object, 10);
 ```
 
-### Null
+### Array of Primitive Types
+The automatic marshalling works only for cases with arrays of objects. In cases where you have a method that takes an array of primitive types, we need to convert as follows:
```
ggarabedian: I recommend using **you** instead of **we** and adding an **it**: 
(...) array of primitive types, you need to convert it as follows: (...)

docs/core-concepts/android-runtime/marshalling/java-to-js.md
```diff
@@ -137,15 +138,48 @@ var jaggedBooleanArray3 = Array.create("[[Z", 10);
...
+```Java
+public static void myMethod(int[] someParam)
+```
+Then we need to invoke it as follows:
```
ggarabedian: Same recommendation as above - **you** instead of **we**. "You" describes more clearly as to who is supposed to take the action - the user.

docs/core-concepts/android-runtime/marshalling/java-to-js.md
```diff
@@ -137,15 +138,48 @@ var jaggedBooleanArray3 = Array.create("[[Z", 10);
...
+SomeObject.myMethod(arr); // assuming the method is accepting an array of primitive types
+```
+
+### Two-Dimensional Arrays of Primitive Types
+
+The above scenario gets more tricky with two-dimensional arrays. Consider a Java method that accepts as an argument a two-dimensional array:
+```Java
+public static void myMethod(java.lang.Integer[][] someParam)
+```
+The JavaScritp marshaled code will look like this:
```
ggarabedian: JavaScri**pt** and marsha**l**led :) 
Also, I am not sure if the meaning will remain the same (to me it does), but "marshaled JavaScript code" sounds better.

ggarabedian (approved):

tsonevn (approved):

PR #1264, opened 2018-07-18
docs: switch iTunes Connect to App Store Connect

renestalder:
iTunes Connect is now App Store Connect. Change references in the documentation to use App Store Connect instead of iTunes Connect.

## PR Checklist

The iOS publishing documentation references to iTunes Connect. While this still works for most of the cases (except the links to the documentations), it can be confusing for new users of NativeScript.

Replace all references of iTunes Connect in the iOS publishing documentation with App Store Connect and change links to new documentation pages.


etabakov (approved):
That's great. Thanks!

Issue #1248, opened 2018-07-06
Create article about complex DTS generation


NickIliev:
The Android DTS generator is now improved and provides more functionality for generating declaration files.

**TODO:**
- Create articles about complex DTS generation  (use [README](https://github.com/NativeScript/android-dts-generator/tree/trifonov/improve-generation)).
- improve[ the article section](https://docs.nativescript.org/core-concepts/accessing-native-apis-with-javascript) for using `tns-platform-declarations` (which are now coming with generated DTS files for all the different API levels)

NickIliev:
Done via [https://github.com/NativeScript/docs/pull/1266](https://github.com/NativeScript/docs/pull/1266)

PR #1266, opened 2018-07-19
docs: improved tns-platform-declarations section + article for Android DTS generator

NickIliev:
Related to [https://github.com/NativeScript/docs/issues/1248](https://github.com/NativeScript/docs/issues/1248)

### DIFF
--- a/docs/core-concepts/accessing-native-apis-with-javascript.md
+++ b/docs/core-concepts/accessing-native-apis-with-javascript.md
@@ -14,7 +14,7 @@ In this article we are going through the basic concepts of how native APIs are a
 
 NativeScript lets you access all native APIs from the underlying platform. To achieve this behaviour, many things happen under the hood. One of them is marshalling - the conversion between JavaScript and Objective-C data types for iOS and Java data types for Android.
 
-In this article, you will learn how to call native APIs from JavaScript with various data types parameters. For more information, see the platform-specific resources about data conversion in the [iOS Runtime](./../runtimes/ios/marshalling/Marshalling-Overview.md) and [Android Runtime](./../runtimes/android/marshalling/overview.md) sections.
+In this article, you will learn how to call native APIs from JavaScript with various data types parameters. For more information, see the platform-specific resources about data conversion in the [iOS Runtime](./ios-runtime/Marshalling-Overview.md) and [Android Runtime](./android-runtime/marshalling/overview.md) sections.
 
 ## Numeric Types
 
--- a/docs/core-concepts/accessing-native-apis-with-javascript.md
+++ b/docs/core-concepts/accessing-native-apis-with-javascript.md
@@ -150,9 +150,9 @@ var button = new android.widget.Button(context);
 button.setOnClickListener(undefined); // the Java call will be made using the null keyword
 ```
 
-## Intellisense and access to native APIs via TypeScript
+## IntelliSense and Access to the Native APIs via TypeScript
 
-To have access and Intellisense for the native APIs with TypeScript enabled project, you have to add a dev dependency to `tns-platform-declarations` 
+To have access and Intellisense for the native APIs, you have to add a dev dependency to `tns-platform-declarations` 
 
 Steps to install and enable 
 
--- a/docs/core-concepts/accessing-native-apis-with-javascript.md
+++ b/docs/core-concepts/accessing-native-apis-with-javascript.md
@@ -158,7 +158,7 @@ Steps to install and enable
 
 - `npm install tns-platform-declarations --save-dev`
 
- > **Note:** Always install the plugin as a `devDependency` (NPM install with `--save-dev` flag) to avoid bringing the enormously big declaration files in the output built file.
+ > **Note:** Always install the plugin as a `devDependency` (`npm i tns-platform-declarations --save-dev` flag) to avoid bringing the enormously big declaration files in the output built file.
 
 - As of version 3.0.0 of NativeScript the newly created projects are shipped without `reference.d.ts` file.
 
@@ -175,7 +175,8 @@ For example, let's assume you are developing an application for API 21+ and you 
  /// <reference path="node_modules/tns-platform-declarations/android-21.d.ts" />
  ```
 
- > **Note:** Proceed with caution when using functionalities introduced in newer API level. If you attempt to use a class, method, or property from a newer API level on a device with a lower API, the application will crash.
+ > **Note:** Proceed with caution when using functionalities introduced in newer API level. If you attemp
+  to use a class, method, or property from a newer API level on a device with a lower API, the application will crash.
 
 - Modify `tsconfig.json` to contain the following settings:
 ```
--- a/docs/core-concepts/accessing-native-apis-with-javascript.md
+++ b/docs/core-concepts/accessing-native-apis-with-javascript.md
@@ -168,14 +168,14 @@ Create `reference.d.ts` in the root project directory and add the following:
 /// <reference path="node_modules/tns-platform-declarations/ios.d.ts" />
 ```
 
- By default, the file `android.d.ts` comes with typings generated for API level 17. As an Android developer, you might need access to a specific class, method or property introduced in a newer API level. The `tns-platform-declarations` plugin comes with generated typings for all API levels from 17 to 27 including the related typings from the respective support library. To use typings for a specific Android level replace the reference to the default declaration file with the preferred one. The files for each API level comes postfixed with a dash followed by the number of the API level (e.g. for API 21 the file is named `android-21.d.ts`).
+ By default, the file `android.d.ts` comes with typings generated for API level 17. As an Android developer, you might need access to a specific class, method, or property introduced in a newer API level. The `tns-platform-declarations` plugin comes with generated typings for all API levels from 17 to 27 including the related typings from the respective support library. To use typings for a specific Android level replace the reference to the default declaration file with the preferred one. The files for each API level comes postfixed with a dash followed by the number of the API level (e.g. for API 21 the file is named `android-21.d.ts`).
 
 For example, let's assume you are developing an application for API 21+ and you need typings generated for that API level:
  ```
  /// <reference path="node_modules/tns-platform-declarations/android-21.d.ts" />
  ```
 
- > **Note:** Proceed with caution when using functionalities introduced in newer API level. If you attempt to use a class, method or property from a newer API level on a device with a lower API, the application will crash.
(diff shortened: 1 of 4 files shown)


VladimirAmiorkov (approved):

docs/core-concepts/accessing-native-apis-with-javascript.md
```diff
@@ -150,22 +150,32 @@ var button = new android.widget.Button(context);
...
 
-To have access and Intellisense for the native APIs with TypeScript enabled project, you have to add a dev dependency to `tns-platform-declarations` 
+To have access and Intellisense for the native APIs, you have to add a dev dependency to `tns-platform-declarations` 
 
 Steps to install and enable 
 
 - `npm install tns-platform-declarations --save-dev`
 
-- As of version 3.0.0 of NativeScript the newly created projects are shipped without `references.d.ts` file.
+ > **Note:** Always install the plugin as a `devDependency` (NPM install with `--save-dev` flag) to avoid bringing the enormously big declaration files in the output built file.
```
etabakov: Why not just post the exact command, instead of explaining "NPM install with `--save-dev` flag?

docs/core-concepts/accessing-native-apis-with-javascript.md
```diff
@@ -150,22 +150,32 @@ var button = new android.widget.Button(context);
...
-Create `references.d.ts` in the root project directory and add the following:
+- As of version 3.0.0 of NativeScript the newly created projects are shipped without `reference.d.ts` file.
+
+Create `reference.d.ts` in the root project directory and add the following:
 ```
 /// <reference path="node_modules/tns-platform-declarations/android.d.ts" />
 /// <reference path="node_modules/tns-platform-declarations/ios.d.ts" />
 ```
 
+ By default, the file `android.d.ts` comes with typings generated for API level 17. As an Android developer, you might need access to a specific class, method or property introduced in a newer API level. The `tns-platform-declarations` plugin comes with generated typings for all API levels from 17 to 27 including the related typings from the respective support library. To use typings for a specific Android level replace the reference to the default declaration file with the preferred one. The files for each API level comes postfixed with a dash followed by the number of the API level (e.g. for API 21 the file is named `android-21.d.ts`).
```
etabakov: Let's add an Oxford comma in "need access to a specific class, method or property"?

docs/core-concepts/accessing-native-apis-with-javascript.md
```diff
@@ -150,22 +150,32 @@ var button = new android.widget.Button(context);
...
+
+For example, let's assume you are developing an application for API 21+ and you need typings generated for that API level:
+ ```
+ /// <reference path="node_modules/tns-platform-declarations/android-21.d.ts" />
+ ```
+
+ > **Note:** Proceed with caution when using functionalities introduced in newer API level. If you attempt to use a class, method or property from a newer API level on a device with a lower API, the application will crash.
```
etabakov: Again, use the Oxford comma in "attempt to use a class, method or property".

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
+---
+nav-title: "Generating TypeScript Declaration Files"
+title: "Generating TypeScript Declaration Files"
+description: "Generating TypeScript declaration files for native Android libraries"
```
etabakov: We can add more elaborative description. Let's summarise why a user would like to read this article, since this description is displayed in the search results of the search engines.

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Generating TypeScript Declarations
+
+## Overview
+
+One of the main advantages of TypeScript is the type checking. TypeScript uses declaration files (a.k.a definitions) to "recognize" the types and function signatures of a given module. In the context of NativeScript mobile development, a developer may need to access the native Android SDK via data conversion (a.k.a. [marshalling](../marshalling/overview.md)) which is converting Java to JavaScript (or vice-versa). When working with TypeScript (or with Angular where the first-class citizen language is also TypeScript), the developer would also need to explicitly declare all namespaces, classes, properties, etc. that are used from the native Java or Objective-C APIs.
+
+### The Bad Practice
+
+When we are using plain JavaScript, all you need is to know the conversion syntax rules, but when TypeScript is used, then each class, method, or property should have a static declaration. We might merely declare all namespaces and variables with the type `any`
```
etabakov: Missing punctuation at the end of the paragraph.

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
...
+let androidContext = android.content.Context;
+```
+
+The above is a bad practice for two reasons:
+- As the namespace is cast to `any` you won't have any intelligence. No matter if working with the Android SDK or third-party Android library having easy access to all properties, methods and classes via IntelliSense is always a great advantage.
+- Risk of runtime errors. Casting to `any` is simply adding the chance to use unexisting functionality (or typing wrong name) and crash the application runtime.
+
+### The Good Practice
+
+To resolve the above, in NativeScript, there is a developer dependency called [tns-platform-declarations.](https://preview.npmjs.com/package/tns-platform-declarations) After the intial instalation and setup, you could directly access the `android` namespace from the Android SDK.
```
etabakov: Let's exclude the period from the "tns-platform-declarations." link.

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
...
+// follow the setup instructions at https://preview.npmjs.com/package/tns-platform-declarations
+let androidContext = android.content.Context;
+```
+
+The `tns-platform-declarations` plugin is providing access to the Android SDK. The plugin comes with pre-generated TypeScript declaration files for all API levels from 17 to 27 exclusive (detailed usage instructions in [this documentation section](../../accessing-native-apis-with-javascript#intellisense-and-access-to-the-native-apis-via-typescript)). In cases, where we need a declaration file for a third-party library or newer API (e.g. API 28) we can generate our own definitions using the **[Android DTS generator](#android-dts-generator)**.
+
+## Android DTS Generator
+
+In cases where you need to generate a TypeScript declaration file for a specific Android SDK or third-party Android library,
+ you can use [Android DTS generator](https://github.com/NativeScript/android-dts-generator/). The Android DTS generator is a tool that generates TypeScript declaration files (`*.d.ts`) from **JAR** or **AAR** libraries.
```
etabakov: Unneeded new line.

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
...
+- Extract the `classes.jar` and any dependencies it may have inside `libs/` folder
+- Rename `classes.jar` if necessary
+- Execute te following command
+```
+java -jar build/libs/dts-generator.jar -input classes.jar dependency-of-classes-jar.jar
+```
+
+## Complex Definitions Generation
+
+Generating the typings corresponding to the android and android-support jar files is a bit tricky operation, so here's a detailed explanation how to do it. There are different andoid support versions and they depend on main android classes it is a little bit complicated to generate those typings. One option is to generate a big d.ts file with all the libraries inside. The downside of this approach is that you have to generate a big file for every API level and if the android support version is changed all those d.ts files need to be regenerated. To avoid this, there's some functionality in the tool for passing dependencies when generating typings. There are two type of dependencies that can be passed:
```
etabakov: - a typo in "andoid" 
- let's also capitalise "android" everywhere we are mentioning it
- the second sentence sounds complicated,

vtrifonov (approved):

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
...
+
+- **Super class jars**- this is needed when the current jar has classes which are extending classes from another jar file, but we don't want to have all that jar files' typings in a single output file. To achieve this, we can provide the super class jar with the super argument(which works the same way as input for multiple files). For instance, if we want to generate typings for **android.support.v4.view.ViewPager** and we don't pass the super classes jar the generated typings won't extend any class as there's no information in the jar that contains the ViewPager. However, this class extends **android.view.ViewGroup** class which is a part of the android jar file(any of the API levels). So if we pass one of the android.jar files as a super class jar file the generated typing will contain `extends android.view.ViewGroup`.
```
etabakov: - let's leave a space in "argument(which" before the opening bracket 
- the same in "file(any of the API levels)."

docs/core-concepts/android-runtime/metadata/generating-typescript-declarations.md
```diff
@@ -0,0 +1,156 @@
...
+./gradlew dependencies --configuration testCompileOnly
+```
+
+- Run the dts-generator tool passing as input arguments the path to the output jars folder
+
+### Android Support Specifics
+
+To get all the jar files for android support library, follow the steps above. You can find the jar files for android support 27.0.1 [in the current repository](https://github.com/NativeScript/android-dts-generator/tree/master/libs/android-support/27.0.1) As the android support needs the base android jar file to create its typings you need to pass the android.jar file as a super parameter to the generator. To avoid having typings for every different API level you can reuse typings built with API level 17 for all API levels until 23. It's quite easy to test this:
+
+1. Run the typings generator for android support passing android-17/android.jar as a supper jar
```
etabakov: - a typo in "supper"?

docs/core-concepts/accessing-native-apis-with-javascript.md
NickIliev: I wanted to explicitly underline this one - in my experience, a lot of the NativeScript's developers are not aware of the difference between `dependency` and `devDependency` and as a common mistake are installing developer plugins as a dependencies (even some of the plugin's authors are installing `tns-platform-declarations` as a dependency)

etabakov: That's great. I think the note is good and it's worth mentioning. I just suggest in the brackets instead of the wording "NPM install with `--save-dev` flag" to post the exact command "npm i tns-platform-declarations --save-dev".

etabakov (approved):

PR #1280, opened 2018-07-29
Update ns-setup-linux.md

DanekDanek:
The previous solution only creates variable for one session. After the restart of the system variable was gone and the whole process was broken. Even after the end of this set up tutorial, during `tns doctor` command, there was a lot of errors about Android SDK. Creating a permanent environment variable fix this and after restart of the system this set up instruction works.

## PR Checklist


sis0k0 (approved):

etabakov (approved):
Thanks for the contribution, @DanekDanek!

PR #1282, opened 2018-07-30
removed internal TOCs from UI sections

NickIliev:
Removed obsolete internal TOCs from several articles (new right TOCs are automatically generated based on the markdown headings). Some minor improvements on-the-fly in the articles.


etabakov (approved):
👍

PR #1303, opened 2018-08-07
new article, describing the new ways for error handling

tsonevn:
new article, describing the new ways for error handling

### DIFF
--- a/docs/core-concepts/error-handling.md
+++ b/docs/core-concepts/error-handling.md
@@ -39,7 +39,7 @@ The example demostrates, how we can use the `error` method inside of a plugin's 
 
 ### Defining custom error handler
 
-This API also introduces a technic, which allows the developers to define custom error handler logic.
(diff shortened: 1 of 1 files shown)


docs/core-concepts/error-handling.md
```diff
@@ -0,0 +1,87 @@
...
+    return;
+  }
+  // ... implementation using arg
+}
+```
+The example demostrates, how we can use the `error` method inside of a plugin's code, inseed of throwing the error, which can cause a crash.
+
+### Defining custom error handler
+
+This API also introduces a technic, which allows the developers to define custom error handler logic.
```
vakrilov: Add: 
"The `ErrorHandler` will be called whenever `trace.error(...)` is called.  The default error handler, which will just `throw` the error as they come."

vakrilov (approved):
Add an extra info about `ErrorHandlers`
Otherwise LGTM

PR #1315, opened 2018-08-18
docs: delete obsolete paragraph

jrpool:
Following the link to the webinar merely produces a page saying the webinar has ended. This makes the paragraph useless. If the webinar were recorded and available, then the paragraph would be useful, but it would need to omit the statement that the user will have an opportunity to ask questions.

## PR Checklist

The page contains an obsolete paragraph.

The obsolete paragraph is no longer present.


etabakov (approved):
Thanks, @jrpool!

PR #1405, opened 2018-10-23
fixing scss snippet

tsonevn:
fixing scss snippet


Fatme (approved):

mikaelkalt:
I think it's still not correct since it's a difference if you're using webpack or not...? 

If you're not using webpack, I think you should still point to the `.css` files and in case of webpack you should use the `.scss` files. - In my opinion it would be worth to mention that.

PR #1410, opened 2018-10-30
fix: Remove the NativeScript Documentation link from Get Started navigation

ggarabedian:
Unless I am missing something, adding the `publish: false` variable to the landing page (landing.md) removes the `NativeScript Documentation` link from the navigation without causing issues.


tsonevn (approved):

etabakov (approved):
I think that we tried this and when `publish: false` is set - the page was not available at all. But if looks ok on your machine - great.

PR #1475, opened 2018-12-14
docs(android-runtime): Document markingMode: "none" option

mbektchiev:
## PR Checklist

The `markingMode: "none" ` option is not documented.

The `markingMode: "none" ` option is now documented.


vtrifonov (approved):

docs/core-concepts/android-runtime/advanced-topics/memory-management.md
```diff
@@ -35,6 +35,25 @@ if the actual `id` has value `123`. In other worlds, creating JavaScript objects
...
+  "android": {
+    "markingMode": "none"
+  }
+}
+```
+
+The code inside `tns-core-modules` and all plugins published by the NativeScript Team (since version 5.1.0) are written in such a way, that it does not depend on the scope to keep those Java instances alive. This makes apps using these plugins fully compatible with the a lot more performant `markingMode: "none"` option.
+
+> **WARNING**: Enabling this option if the JavaScript code dealing with native objects (either in the application or any plugins you are using) does not correctly take care of the lifetime of Java instances may cause unexpected and unpredictable crashes of the application due to Java instances being prematurely collected. Use caution when enabling it and make sure to thoroughly test your apps with different memory constrains and devices!
+
```
etabakov, line 57: Should we provide information about the specific faulty behaviour that the users should be careful? A specific exception, crashes, etc.

mbektchiev: Good idea, I'll include the error message.

etabakov (approved):

PR #1483, opened 2018-12-18
png/gif image optimization

rdlauer:
Using tinypng and ezgif, optimized all png/gif files in the docs (basically lossless optimization, so there shouldn't be any noticable difference in quality).

Total went from **32.6MB** to **13.9MB**.


NickIliev (approved):

etabakov:
Awesome!

PR #1488, opened 2018-12-26
Update ns-setup-linux.md

jofftiquez:
As discussed in [this SO question](https://stackoverflow.com/questions/7535569/what-is-an-appropriate-directory-in-which-to-install-android-sdk). `/usr/local` should be the right place for the android sdk.

Also putting it is the `/` seems to be a bad idea. The android sdk is too large to be put in `/` in my opinion.

Regarding the environmental variable, I user `~/.bashrc` in favor of `/etc/environment`. For some reason using `/etc/environment` won't cut it. 

A [reference](https://stackoverflow.com/questions/26256279/how-to-set-android-home-path-in-ubuntu) for using `~/.bashrc`

## PR Checklist

### DIFF
--- a/docs/start/ns-setup-linux.md
+++ b/docs/start/ns-setup-linux.md
@@ -63,21 +63,20 @@ Complete the following steps to set up NativeScript on your Linux development ma
         <pre class="add-copy-button"><code class="language-terminal">sudo update-alternatives --config java
         </code></pre>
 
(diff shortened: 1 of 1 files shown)


NickIliev (approved):

jofftiquez:
Thank you @NickIliev

PR #1492, opened 2019-01-02
Fix scroll sidebar docs

### LINKED ISSUE #1455: Can't scroll "In this Article" column
### Please, tell us what's the problem?
[ ] A typo
[ ] Wrong documentation
[x] Improvement of existing article
[ ] Missing documentation
[ ] New article needed

### Please, tell us more details
I can't scroll the right column ("In This Article") separately. I have to scroll the whole page to the end to see if there is more links (headings), which doesn't benefit me at that point. Same is true for the left column (Topics). For example, open [The Basics page](https://docs.nativescript.org/ui/basics).

bdlb77:
## PR Checklist

- [x ] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x ] If there is an issue related with this PR, point it out here.

   #1455 , implemented scroll effect for both left and right sidebars. Both sidebars didn't have the ability to scroll on content, causing difficulty in UX and navigation on tabs. 

  The side navbars should have the ability to scroll to allow for easier access to tabs. Also they are now sticky on the page to always show and have access to their navigations.

Fixes #1455


etabakov:
Many thanks for your effort. While reviewing the suggestion I noticed a redundant scrollbar, that ideally shouldn't be there - check the screenshot.

bdlb77:
I found a solution for the navbar that possibly only works for webkit and firefox.. I'm not sure how it looks on IE... maybe y'all have a suggestion on how to implement a fix for it there too :) Thank you guys for this opportunity! 
screen shot 2019-01-03 at 2 58 16 pm
screen shot 2019-01-03 at 2 58 23 pm

build/_assets/javascripts/app.js
```diff
@@ -362,7 +362,7 @@ $(function(){
     var rightNavLinks = $(".right-nav__links");
 
     var rightNav = $('\
-<div class="right-nav__container">\
+<div class="right-nav__container navigation__right__scroll">\
```
bundyo: You don't really need a special class for your CSS, you can just add it to ```.right-nav_container```.

build/_assets/javascripts/app.js
```diff
@@ -497,3 +497,43 @@ $(function() {
...
+
+setTimeout(NsSideScroll, 1500);
+
+// Handle Navigation Right Scroll 
+
+function handleScrollNav(){
+    const rightNavScroll = document.querySelector('.right-nav__container');
+    const pageContent = document.querySelector('#page-article article');
+    if(pageContent){
+        function handleRightScroll(e){
```
bundyo: This function doesn't seem needed, as you call it only once immediately after its definition.

build/_assets/javascripts/app.js
```diff
@@ -497,3 +497,43 @@ $(function() {
...
+
+    window.addEventListener('scroll', handleSticky);
+    
+}
+
+setTimeout(NsSideScroll, 1500);
+
+// Handle Navigation Right Scroll 
+
+function handleScrollNav(){
```
bundyo: The container is already sticky, not sure we need this whole implementation at all if it is done only for the transition. I would advise using only CSS - check my next comments.

build/_assets/javascripts/app.js
```diff
@@ -497,3 +497,43 @@ $(function() {
     window.setTimeout(handleBanner, 1000);
 });
 
+// Handle ns-side-scroll
```
bundyo: I've discussed with our designer if the left side navigation should be scrollable, and we agreed that since it is part of the flow of the page and scrolls with it, this is not necessary. In addition, in the 1280 media breakpoint the menu content scrolls under the menu trigger arrow. So we should remove the scrolling of the left side navigation.

build/_assets/stylesheets/styles.css
```diff
@@ -2770,3 +2790,21 @@ html .gsc-result .gs-title {
         flex-direction: column;
     }
 }
+
+/* Styling for the NS Side navigation scroll */
```
bundyo: Remove these as well.

build/_layouts/page.html
```diff
@@ -4,7 +4,7 @@
 <div class="ns-page">
     <div class="container">
         <div id="page-inner-content" class="row">
-            <div class="ns-side-nav" {{site.sideBarHidden}}>
+            <div class="ns-side-nav"  id="ns-side-scroll" {{site.sideBarHidden}}>
```
bundyo: I didn't really get why did you use the ID (except for styling, which you can do through the class).

build/_assets/stylesheets/styles.css
```diff
@@ -2770,3 +2790,21 @@ html .gsc-result .gs-title {
...
+@media only screen and (min-width: 1024px){
+    #ns-side-scroll{
+        position: sticky !important;
```
bundyo: Avoid using important when you don't have the need to override CSS set with JS.

build/_assets/stylesheets/styles.css
```diff
@@ -780,13 +780,33 @@ html body .side-nav__tree {
...
 .right-nav__container {
+    transition: all 0.5s ease;
     top: 6.25em;
     margin-left: 1.25em;
     position: -webkit-sticky;
     position: sticky;
     float: right;
     margin-right: -17em;
 }
+.navigation__right__scroll {
```
bundyo: After our designer saw your PR in action, we discussed how the feature should look like and decided that the scrollbar should be visible and use the **--border-light-blue color**.

No need for this class - add the overflow-y to the **.right-nav_tree** (which is used for scrolling in the 1280 breakpoint), though better use auto, to avoid showing a scrollbar when it is not needed. Instead of position: fixed, right and and bottom, add just height - something like **calc(100vh - 12.76em)** should be just right for when the yellow header is not visible and the page is not scrolled. Add also scrollbar-width: thin (instead of none) and scrollbar-color to be transparent.

Then, on hover change the scrollbar-color to the one I mentioned above. This should take care of Firefox.

I'll leave the Webkit/Chrome styling to you - make it 6px wide, transparent and the same color on hover.

Make sure to change the height when the yellow bar is visible and when the page is scrolled (the header is detached).

Also check the 1280px media breakpoint and make sure the height is still 100% when you resize to it.

bundyo (changes requested):
I hope the changes I'm requesting are not too much. 

Maybe we should have provided a design requirement beforehand. Sorry for missing that.

build/_assets/javascripts/app.js
bdlb77: 👍

build/_assets/javascripts/app.js
bdlb77: 👍

build/_layouts/page.html
bdlb77: true true, I took it out!

build/_assets/stylesheets/styles.css
bdlb77: turned out not to be a need with this styling

build/_assets/stylesheets/styles.css
bdlb77: So yes you were right.. no JS needed haha..  I made this waaay too complicated than it needed to be. Thank you for the great advice! 
I'll add photos for the updated version. 

screen shot 2019-01-09 at 2 10 34 pm
screen shot 2019-01-09 at 2 10 15 pm
screen shot 2019-01-09 at 2 08 30 pm

bdlb77:
So yes you were right.. no JS needed haha..  I made this waaay too complicated than it needed to be. Thank you for the great advice! 
I'll add photos for the updated version. 

screen shot 2019-01-09 at 2 21 45 pm
screen shot 2019-01-09 at 2 21 37 pm
screen shot 2019-01-09 at 2 21 25 pm


Once I do a full check again I'll push the changes

bundyo:
Thanks. 

Btw, looking at your screenshots - the colors of the scrollbar are inverted than what I had in mind - the scroll-thumb should be --border-light-blue and the scroll-track should be transparent.

bdlb77:
screen shot 2019-01-09 at 3 07 49 pm
Hopefully this one will be what you were lookin for :) sorry for the confusion

bdlb77:
I just pushed the next round of revisions :) gracias

bundyo (approved):

bundyo:
Thanks for your effort! :+1:

bdlb77:
Thank you guys! I learned lots :)

etabakov:

Regards from the whole team 🥇

PR #1495, opened 2019-01-08
docs(core-concepts): re-work angular navigation article

MartoYankov:


docs/core-concepts/angular-navigation.md
```diff
@@ -3,19 +3,113 @@ title: Navigation
...
+
+* Native navigation transitions.
+* Back navigation handling - the hardware back button on Android and the navigation bar back button on iOS.
+* Mobile navigation lifecycle - view state preservation when navigating back.
+* Mobile specific history - keeping history per navigation controller instead of a global one.
+* Mobile lateral navigation widgets - `TabView`, `SideDrawer`, `Modal View` and more.
+
+NativeScript brings these to Angular with the following extensions, directives and strategies:
+
+* `page-router-outlet` - This is an alternative to the regular `router-outlet` that serves as a placeholder for where native mobile navigation will occur.
```
etabakov: In the bullet list above we use lower case after the dash, and in this a capital letter is used. Let's stick to one of them, I think we use lower case in most of the places.

docs/core-concepts/angular-navigation.md
```diff
@@ -3,19 +3,113 @@ title: Navigation
...
+* Custom [``PlatformLocation``](https://angular.io/api/common/PlatformLocation) and [``LocationStrategy``](https://angular.io/api/common/LocationStrategy) - This strategy keeps history per outlet instead of one global linear history.
+
+We will explore each of these in the following sections.
+
+### Page Router Outlet
+
+NativeScript's `page-router-outlet` is the equivalent of Angular's [``router-outlet``](https://angular.io/api/router/RouterOutlet). It serves as a placeholder for native mobile navigations. Internally, each `page-router-outlet` creates a NativeScript `Frame` and each component that the router displays in the outlet is wrapped in a `Page` widget. This is the main integration point that brings native navigation to Angular. The `Frame` and `Page` combination also means you can use the `ActionBar` widget in these components. For more information on how NativeScript Core navigation works visit this [documentation article]({% slug navigation %}.
```
etabakov: Broken link

```suggestion
NativeScript's `page-router-outlet` is the equivalent of Angular's [``router-outlet``](https://angular.io/api/router/RouterOutlet). It serves as a placeholder for native mobile navigations. Internally, each `page-router-outlet` creates a NativeScript `Frame` and each component that the router displays in the outlet is wrapped in a `Page` widget. This is the main integration point that brings native navigation to Angular. The `Frame` and `Page` combination also means you can use the `ActionBar` widget in these components. For more information on how NativeScript Core navigation works visit this [documentation article]({% slug navigation %}).
```

docs/core-concepts/angular-navigation.md
```diff
@@ -3,19 +3,113 @@ title: Navigation
...
+export class MainComponent {
+
+	constructor(private routerExtensions: RouterExtensions) {
+	}
+}
+```
+
+Here is a list of the available methods:
+
+* `navigate()` - This method is an alternative to the Angular `Router` [``navigate()``](https://angular.io/api/router/Router#navigate) method, but also supports navigations in a `page-router-outlet`.
```
etabakov: Same as the first comment about the capitalisation.

docs/core-concepts/angular-navigation.md
```diff
@@ -3,19 +3,113 @@ title: Navigation
...
+
+![router-outlet-lifecycle](../img/navigation/router-outlet-lifecycle.png?raw=true)
+
+In a native mobile application the system will keep the navigated views alive, so that when you come back to them, their view state will be kept the same. Views are destroyed only when you back away from them. The `page-router-outlet` houses native navigations, so its components lifecycle must match the lifecycle of the native views. This is done by the custom `NSRouteReuseStrategy`.
+
+![page-router-outlet-lifecycle](../img/navigation/page-router-outlet-lifecycle.png?raw=true)
+
+You might want to perform some cleanup actions (e.g. unsubscribe from a service to stop updates) when you are navigating forward to a next page. If you are using `page-router-outlet` you cannot do that in the `ngOnDestroy()` hook, as this will not be called when you navigate forward. What you can do is inject the `Page` instance inside your component and attach to page navigation events (for example `navigatedFrom`) and do the cleanup there. You can check all the available page events [here](https://docs.nativescript.org/ui/ns-ui-widgets/page#page-events).
```
etabakov: Not sure if 'to' is required here, but sounds better to me at least. Feel free to disregard.

```suggestion
You might want to perform some cleanup actions (e.g. unsubscribe from a service to stop updates) when you are navigating forward to a next page. If you are using `page-router-outlet` you cannot do that in the `ngOnDestroy()` hook, as this will not be called when you navigate forward. What you can do is to inject the `Page` instance inside your component and attach to page navigation events (for example `navigatedFrom`) and do the cleanup there. You can check all the available page events [here](https://docs.nativescript.org/ui/ns-ui-widgets/page#page-events).
```

docs/core-concepts/angular-navigation.md
```diff
@@ -30,176 +124,1181 @@ Use the `NativeScriptRouterModule` API to import your routes:
...
+The combination of NativeScript and Angular is very powerful in terms of navigation options. Angular brings its own well known navigation mechanism with a router. NativeScript on the other hand provides access to the native mobile navigation patterns. Due to the nature of the integration between the two, you have a choice to use either of them or even a combination that suits your needs. In the following sections we will provide implementation examples to various mobile navigation patterns.
 
-> Note: You will rarely need to create Page instances manually. The framework creates pages automatically when bootstrapping or navigating the app. You can get a reference to the current page by injecting it into your component using the DI.
+>**Note:** To improve the bootstrap time and the in-app navigation in large applications, Angular has introduced **lazy loading** for modules. Refer to the [Lazy Loading in NativeScript](../performance-optimizations/lazy-loading.md) for detailed explanation and implementation steps. We won't use this feature in the following examples for the sake of simplicity.
```
etabakov: Remove file extention in the hyperlink to avoid a redundant redirect:

```suggestion
>**Note:** To improve the bootstrap time and the in-app navigation in large applications, Angular has introduced **lazy loading** for modules. Refer to the [Lazy Loading in NativeScript](../performance-optimizations/lazy-loading) for detailed explanation and implementation steps. We won't use this feature in the following examples for the sake of simplicity.
```

etabakov (approved):
Love it! Clear illustration and in-depth explanation of the problem as a whole. Some minor tweaks from me.

docs/core-concepts/angular-navigation.md
```diff
@@ -3,19 +3,113 @@ title: Navigation
...
+
+We recommend that you use `page-router-outlet` for your major mobile navigation pattern and use the regular `router-outlet` for internal component navigations if needed. You are also free to use only the `router-outlet` if this makes more sense for your scenario.
+
+### Router Link
+
+In a NativeScript application you **cannot** use the Angular [`routerLink`](https://angular.io/api/router/RouterLink) directive. NativeScript provides its own `nsRouterLink` directive that is working much in the same way. Additionally, it supports two NativeScript specific attributes that you can add to your `nsRouterLink` tag in the markup.
+
+* pageTransition - This attribute lets you specify the native transition for the `nsRouterLink` navigation. Accepted values are `true`, `false`, one of the predefined transitions listed [here](https://docs.nativescript.org/api-reference/interfaces/_ui_frame_.navigationtransition#name) or a custom [NavigationTransition](https://docs.nativescript.org/api-reference/interfaces/_ui_frame_.navigationtransition) object.
```
vakrilov, line 49: Also add `pageTransitionDuration`- https://github.com/NativeScript/nativescript-angular/blob/master/nativescript-angular/router/ns-router-link.ts#L51

tsonevn (approved):

docs/core-concepts/angular-navigation.md
```diff
@@ -3,19 +3,113 @@ title: Navigation
...
+
+### Passing Parameter
+
+In Angular you can inject `ActivatedRoute` and read route parameters from it. Your component will be reused if you do a subsequent navigation to the same route while only changing the params. That's why `params` and `data` inside `ActivatedRoute` are observables. Using `ActivatedRoute` is covered in [angular route-parameters guide](https://angular.io/api/router/ActivatedRoute).
+
+As explained in the previous section, with `page-router-outlet` when navigating **back** to an existing page, your component will **not** be re-created. However, Angular router will still create a **new instance** of `ActivatedRoute` and put all params in it. You cannot get hold of it through injection, as your component is revived from the cache and not constructed anew.
+
+**The Solution**: In NativeScript you can inject `PageRoute` which has an `activatedRoute: Observable<ActivatedRoute>` field inside it. Each time a new `ActivatedRoute` instance is created for this reused component it will be pushed in this observable, so you can still get your params:
```
vakrilov: I think this part is severely outdated - `PageRoute` was introduced before we started using a custom route-reuse-strategy. It should not be needed now as the reuse strategy _should_ handle `activatedRoutes` correctly. 

Let's double check that and remove the reference to `PageRoute` if it not needed.

vakrilov (changes requested):

docs/core-concepts/angular-navigation.md
vakrilov: Yup - everything works with `ActivatedRoute` as expected. We can remove the whole `Passing Parameter` section from the docs as there is nothing NativeScript specific.

vakrilov (approved):
👍 Once comments are resolved.

docs/core-concepts/angular-navigation.md
MartoYankov: I intentionally forgot to mention it :) The reason is you can set a whole `NavigationTransition` object (which contains duration) to the `pageTransition` attribute.

PR #1502, opened 2019-01-14
breakout scss vars into their own file

rynop:
I was having lots of trouble getting [icon fonts](https://docs.nativescript.org/ui/icon-fonts) to work inside a `<Button>` when applying the `btn` class.  Ex:

``` html
<Stacklayout>
    <!-- Label always renders icon font -->
    <Label class="fas" text="{{'&#xf008;'}} film"></Label>
    <!-- Button with no btn class always works -->
    <Button [nsRouterLink]="['../showtime']" class="fas btn-primary" text="{{'&#xf008;'}} film"></Button>
    <!-- btn class changes font-family, breaking rendering of icon font (renders ?) -->
    <Button [nsRouterLink]="['../showtime']" class="fas btn btn-primary" text="{{'&#xf008;'}} film"></Button>
</Stacklayout>
```

Before this PR, adding the `btn` to the `class=""` attribute breaks the rendering of font-icon fonts.  I think this is due to the inheritance of the `font-family`.

IMO breaking out app specific SCSS vars into their own file (`_app-variables.scss`) is cleaner, and as a bonus it fixes the issue I was running into, by changing the order of scss loading.

For completeness sake, here is my `_app-common.scss`

``` css
// Main theme ruleset
@import '~nativescript-theme-core/scss/index';
// Place any CSS rules you want to apply on both iOS and Android here.
// This is where the vast majority of your CSS code goes.
.far {
    font-family: 'Font Awesome 5 Free', fa-regular-400;
}

.fab {
    font-family: 'Font Awesome 5 Brands', fa-brands-400;
}

.fas {
    font-family: 'Font Awesome 5 Free Solid', fa-solid-900;
}
```

### DIFF
--- a/docs/ui/theme.md
+++ b/docs/ui/theme.md
@@ -700,6 +700,7 @@ With SASS set up and ready to use, next you’ll need to import the theme’s `.
 ```
 .
 ├── _app-common.scss
+├── _app-variables.scss
 ├── app.android.scss
 └── app.ios.scss
 ```
@@ -707,8 +708,12 @@ With SASS set up and ready to use, next you’ll need to import the theme’s `.
 After that, paste the following code into your `app.android.scss` file.
 
 ``` CSS
-@import 'app-common';
+// Import app variables
+@import 'app-variables';
 @import '~nativescript-theme-core/scss/platforms/index.android';
+
+// Import common styles
+@import 'app-common';
 
 // Place any CSS rules you want to apply only on Android here
 ```
@@ -718,22 +723,28 @@ After that, paste the following code into your `app.android.scss` file.
 And the following code into your `app.ios.scss` file.
 
 ``` CSS
-@import 'app-common';
+// Import app variables
+@import 'app-variables';
 @import '~nativescript-theme-core/scss/platforms/index.ios';
 
+// Import common styles
+@import 'app-common';
+
 // Place any CSS rules you want to apply only on iOS here
+```
+
+Create `_app-variables.scss` with the following contents:
+
+``` CSS
+// Import the theme’s variables
+@import '~nativescript-theme-core/scss/light';
+
+// Customize any of the theme’s variables here, for instance $btn-color: red;
 ```
 
 Finally, paste the following code into your `_app-common.scss` file.
 
 ``` CSS
-// Import the theme’s variables. If you’re using a color scheme
-// other than “light”, switch the path to the alternative scheme,
-// for example '~nativescript-theme-core/scss/dark'.
-@import '~nativescript-theme-core/scss/light';
-
-// Customize any of the theme’s variables here, for instance $btn-color: red;
-
 // Import the theme’s main ruleset.
 @import '~nativescript-theme-core/scss/index';
 
--- a/docs/ui/theme.md
+++ b/docs/ui/theme.md
@@ -692,7 +692,7 @@ The NativeScript core theme is written in SASS, and you can (optionally) use the
 To get started, first verify that your app has the [NativeScript SASS plugin](https://github.com/toddanglin/nativescript-dev-sass) installed by running the following command:
 
 ```
-tns install sass
+npm install nativescript-dev-sass --save-dev
 ```
 
 With SASS set up and ready to use, next you’ll need to import the theme’s `.scss` files into your own. Start by creating the following files in your app:
@@ -707,9 +707,12 @@ With SASS set up and ready to use, next you’ll need to import the theme’s `.
 
 After that, paste the following code into your `app.android.scss` file.
 
-``` CSS
+``` SCSS
 // Import app variables
 @import 'app-variables';
+
+// Import the theme’s main ruleset - both index and platform specific.
+@import '~nativescript-theme-core/scss/index';
 @import '~nativescript-theme-core/scss/platforms/index.android';
 
 // Import common styles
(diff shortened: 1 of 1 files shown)


MartoYankov:
@rynop This part of the docs needed an update, so thanks for that. I have only one consideration. I think it would be better to import the theme's main ruleset before the platform specific ruleset. Take a look at how it's done in our templates [here](https://github.com/NativeScript/template-master-detail-ng/tree/master/src).

Basically, move the import
```
@import '~nativescript-theme-core/scss/index';
```

to the `app.android.scss/app.ios.scss` files. It will leave the `_app.common.scss` with only the rules.

rynop:
Yes that is great. Thank you.  A few things:

- The angular demo app in the [nativescript-dev-sass](https://github.com/NativeScript/nativescript-dev-sass/tree/master/demo-angular/app) should be updated as well to be consistent.  If you are OK with this, I can PR this work on that repo. Essentially:
```
rm https://github.com/NativeScript/nativescript-dev-sass/tree/master/demo-angular/app/*.scss
rm -r https://github.com/NativeScript/nativescript-dev-sass/tree/master/demo-angular/app/styles
cp https://github.com/NativeScript/template-master-detail-ng/tree/master/src/*.scss
https://github.com/NativeScript/nativescript-dev-sass/tree/master/demo-angular/app/
```
-  I figured out how to correctly implement multiple `font-family`s on `<Button>` text (one for icon and one for button text).  This problem was what orig led me to open this issue.  Solution is to use `FormattedString`.  Details [here](https://stackoverflow.com/questions/54239580/how-to-use-a-fonticon-inside-a-nativescript-button/54239600#54239600) for others that run across this.

- The [icon-fonts doc article](https://docs.nativescript.org/ui/icon-fonts) needs to be re-written.  The instructions do not work for recent version of Font Awesome (5+) and it would be nice to add how to use an icon within a `<Button>`.  I've written a [blog post](https://rynop.com/2018/12/27/get-font-awesome-5-working-with-nativescript-angular-mobile-app/) that covers all this.  I don't have time ATM to re-write the [icon-fonts doc article](https://docs.nativescript.org/ui/icon-fonts).  I'm including this bullet here in case you are OK with linking to my blog OR if someone has the time to update the doc referencing the content in my blog post.

MartoYankov:
@rynop Thanks for the suggestions and the awesome blog post. Regarding this PR, will you update the PR with the @import moved as per my previous comment? I can do this too. 

Regarding your suggestions:
* It's not as simple as running those commands. The master detail template has some specific styles that should be removed. It won't be cool to have styles for elements that don't exist. If you want to make the PR and remove the unneeded styles, go ahead.
* This is a cool idea, perhaps we should add an example with `Button` and `FormattedString` in the UI Widgets section.
* I went through the article and your blog post and I'm not sure that I understand why the instructions in the article don't work for newer versions. Can you elaborate more on that? What doesn't work?

rynop:
-  Can you move the @import? I'm leaning towards the code in this PR not really being useful after seeing https://github.com/NativeScript/template-master-detail-ng/tree/master/src .The SCSS file layout in this repo is exactly what I want.
-  Yup not as easy as those cp cmds, just pseudo code. I'll make a PR against that repo.
-  Re blog post: What doesn't work is under the section titled "The Problem".  I'll try to clarify a bit here:

The NativeScript [icon-fonts](https://docs.nativescript.org/ui/icon-fonts) doc has this snippet at the bottom:

```
/* only for reference, is NOT working in this example (wrong codes) 
.fa {
   font-family: 'FontAwesome', 'fontawesome-webfont';
   font-size: 48;
}
*/
```

This does not work.  A few reasons why:

1.  FontAwesome's `font-family` is no longer `FontAwesome`.  They have [2 families](https://github.com/FortAwesome/Font-Awesome/blob/34dff97269eb4d4142218af6a9dbc44024524ba2/css/all.css#L4197): `Font Awesome 5 Free` and `Font Awesome 5 Brands`.  

2. To make things more complicated, they use the font `Style` within the `Font Awesome 5 Free` family  to differentiate `Regular` (`fa`) and `Solid` (`fas`):

image

image

The `Style` (AFAIK) can't be used by nativescript to tell the native app what `Style` within the `font-family` to use.  Said another way, I can't do something like this:

```
.fa {
   font-family: 'Font Awesome 5 Free', fa-regular-400;
   font-style: Regular;
}
```

Because of this, you need to break out the `Solid` and `Regular` fonts, into their own `font-family`s (per my blog post).

To be completely honest, I haven't tried the `font-style` test above, I was not even aware it was a CSS property.  If it works, the docs still need to be updated, just don't have to run the `python fontname.py` part in my blog post.

I know that was alot of reading - does that make sense?

rynop:
I just confirmed `font-style` can not be used to convey this info to the native app font handling.  For example this does **not** work:

```css
.far {
    font-family: 'Font Awesome 5 Free', fa-regular-400;
    font-style: 'Regular';
}
.fas {
    font-family: 'Font Awesome 5 Free', fa-solid-900;
    font-style: 'Solid';
}
```

nor does this:

```css
.far {
    font-family: 'Font Awesome 5 Free', fa-regular-400;
    font-style: Regular;
}
.fas {
    font-family: 'Font Awesome 5 Free', fa-solid-900;
    font-style: Solid;
}
```

MartoYankov:
@rynop Thanks, this was quite comprehensive. We will investigate this further and decide how to move forward.

In the meantime I will try to update this PR. I think I can do this via your fork, so that your contribution is not lost. Be sure to ping me in the PR in the SASS plugin repo.

rynop:
Do whatever is easiest - no need to do so in my fork.  That is nice of you, but I do not require the recognition, you can close this PR all together if thats easiest.

MartoYankov (approved):

PR #1512, opened 2019-01-21
correct strings.xml path

rynop:
needed for setting app name.  Getting this right is crucial.

### DIFF
--- a/docs/tooling/publishing/publishing-android-apps.md
+++ b/docs/tooling/publishing/publishing-android-apps.md
@@ -46,8 +46,10 @@ You may need to delete `platforms/android` and rebuild using the CLI command `tn
 
 ## App name
 This is the display name for your app. It is purely cosmetic but highly important. For example, it appears under the app icon.
-The value can be stored in the `app/App_Resources/Android/values/strings.xml` file
+The value is set via `App_Resources/Android/src/main/res/values/strings.xml` file.  Create this file if it does not exist.
+
 ```XML
+<?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="app_name">MyAppName</string>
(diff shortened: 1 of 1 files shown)


NickIliev (changes requested):
@rynop thank you for your contribution.

I would like to note that the app name will be set by default to the name given by the user during the creation of the project (related code [here](https://github.com/NativeScript/nativescript-cli/blob/f83c95e6fb4e99911f5036a9795e4dc6bb901cff/lib/services/android-project-service.ts#L273-L274)). However adding your own  `strings.xml` will overwrite the automatically created one which will as a result will require for the user to set both keys manually.

The user can still get the originally set app name with the global variables
```
<string name="app_name">__NAME__</string>
<string name="title_activity_kimera">__TITLE_ACTIVITY__</string>
```

I think It would be best to change `The value is set via the string.xml file.  Create this file if it does not exist.` to something more like 
```
The value can be set via the string.xml file. 
Creating your own strings.xml file will  explicitly require the creation of the following two values

<?xml version="1.0" encoding="utf-8"?>
<resources>
<string name="app_name">MyAppName</string>
<string name="title_activity_kimera">MyAppName</string>
```


More about this [here](https://github.com/NativeScript/nativescript-cli/issues/3670#issuecomment-395207216)

rynop:
Hi Nick.  I'm totally fine with your proposed wording. I'll make the update to this PR.

The most important part of this issue IMO is to update the path to `strings.xml` because as it sits now, it is incorrect.  It is also unclear (was at least to me) that `App_Resources/Android/src/main/res/values/strings.xml` needed to be created if DNE.

NickIliev (approved):

PR #1524, opened 2019-02-01
feature: add docs for markingMode:none

tbozhikov:
A new article on markingMode:none usage

### DIFF
--- a/docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
+++ b/docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
@@ -1,21 +1,23 @@
 ---
-title: "markingMode:none"
-description: "Things to consider when using markingMode:none in your NativeScript mobile app"
+title: "Optimizing Performance Using markingMode: none"
+description: "How to enable markingMode:none in your NativeScript mobile app and test for potential problems"
 position: 4
 slug: marking-mode-none
 ---
 
 > **WARNING**:  Use caution when enabling this option and make sure to thoroughly test your apps with different memory constrains and devices! In this mode NativeScript does not take care of the lifetime of Java instances and unexpected and unpredictable crashes can occur due to Java instances being prematurely collected. The errors generated in such cases look like this:
 `Error: com.tns.NativeScriptException: Attempt to use cleared object reference id=<some-object-id-number>`
+... or like:
+`The JavaScript instance no longer has available Java instance counterpart`
 
-# Introduction
 
-Starting with NativeScript 3.2, a new (experimental at the time) feature was added to the Android runtime called `markingMode`. Its purpose is to speed up garbage collection in the V8 engine. In some cases, a GC pass could take from 0.5 to 1 second and since it runs on the main UI thread, the user would experience a frozen app until GC is done. Setting “markingMode” to “none” will speed up the garbage collection greatly, so it would be less noticeable (if at all) to the app user. The downside of this approach is that some JavaScript objects could be collected while they are still in use. This could happen when JavaScript objects are referenced only from native code and in the eyes of the V8 GC no JS object holds reference to them. In other words – the objects are no longer “marked” as used and the V8 GC might collect them too early.
+# Optimizing Performance Using markingMode: none
+
+Starting with NativeScript 3.2, a new (experimental at the time) feature was added to the Android runtime called `markingMode`. Its purpose is to speed up garbage collection in the V8 engine. In some cases, a GC pass could take from 0.5 to 1 second and since it runs on the main UI thread, the user would experience a frozen app until GC is done. Setting “markingMode” to “none” will speed up the garbage collection greatly, so it would be less noticeable (if at all) to the app user. The downside of this approach is that some objects could be collected while their counterparts (in V8 or Android) are still in use. Such case is when JavaScript objects are referenced only from native code and in the eyes of the V8 GC no JS object holds reference to them. In other words – the objects are no longer “marked” as used and the V8 GC might collect them too early.
 
 The code inside `tns-core-modules` and all plugins published by the NativeScript Team (since version 5.1.0) are written in such a way, that it does not depend on the scope to keep those Java instances alive. This makes apps using these plugins fully compatible with the much more performant `markingMode: "none"` option.
 
-
-## Benefits and drawbacks of setting markingMode to none
+## Benefits and drawbacks of using markingMode: none
 
 The biggest benefit of setting `markingMode` to `none` is a more responsive app – an app that does not slow down if you use it for an extended amount of time. 
 The main drawbacks are:
@@ -37,7 +39,7 @@ If the app behaves correctly after this change - great. If, however, some sporad
 
 ```javascript
 var implementor = new android.native.Implementor();     // native class
-  var callback = new android.native.NCallback({         // native interface
+var callback = new android.native.NCallback({           // native interface
     getMessage: function () {
         implementor.getMessage();
     }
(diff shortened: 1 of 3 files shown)


docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+title: "markingMode:none"
+description: "Things to consider when using markingMode:none in your NativeScript mobile app"
+position: 4
+slug: marking-mode-none
+---
+
+> **WARNING**:  Use caution when enabling this option and make sure to thoroughly test your apps with different memory constrains and devices! In this mode NativeScript does not take care of the lifetime of Java instances and unexpected and unpredictable crashes can occur due to Java instances being prematurely collected. The errors generated in such cases look like this:
+`Error: com.tns.NativeScriptException: Attempt to use cleared object reference id=<some-object-id-number>`
+
+# Introduction
```
etabakov:
```suggestion
## Introduction
```

There should be only one h1 heading and it should be the title of the article.

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
+---
+title: "markingMode:none"
+description: "Things to consider when using markingMode:none in your NativeScript mobile app"
+position: 4
+slug: marking-mode-none
+---
+
```
etabakov, line 7: We should put the title of the article here. Not sure if just "markingMode: none" would be enough.

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+`Error: com.tns.NativeScriptException: Attempt to use cleared object reference id=<some-object-id-number>`
+
+# Introduction
+
+Starting with NativeScript 3.2, a new (experimental at the time) feature was added to the Android runtime called `markingMode`. Its purpose is to speed up garbage collection in the V8 engine. In some cases, a GC pass could take from 0.5 to 1 second and since it runs on the main UI thread, the user would experience a frozen app until GC is done. Setting “markingMode” to “none” will speed up the garbage collection greatly, so it would be less noticeable (if at all) to the app user. The downside of this is that some JavaScript objects would be garbage collected while still in use, because they are referenced only from native code and the eyes of the V8 GC, no JS object holds reference to them. In other words – the objects are no longer “marked” as used and the V8 GC might collect them too early.
```
elena-p: We could split `The downside of this is that some JavaScript objects would be garbage collected while still in use, because they are referenced only from native code and the eyes of the V8 GC, no JS object holds reference to them.` into 2 sentences `The downside of this approach is that some JavaScript objects could be collected while they are still in use. This could happen when JavaScript objects are references only from native code and in the eyes of the V8 GC no JS object holds reference to them.`

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+
+The code inside `tns-core-modules` and all plugins published by the NativeScript Team (since version 5.1.0) are written in such a way, that it does not depend on the scope to keep those Java instances alive. This makes apps using these plugins fully compatible with the much more performant `markingMode: "none"` option.
+
+
+## Benefits and drawbacks of setting markingMode to none
+
+The biggest benefit of setting `markingMode` to `none` is a more responsive app – an app that does not slow down if you use it for an extended amount of time. 
+The main drawbacks are:
+- It is up to the plugin developer to manage the plugin-related memory correctly. Not all plugins support the feature and might crash the app with an "cleared reference" exception.
```
elena-p: `Not all plugins support` -> `There might be plugins, that do not support`
`and might crash` -> `and could crash`

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+    }
+});
+android.native.Executor.printWithDelay(callback, 3s);
+```
+
+The implementor is enclosed by the callback implementation, but with `markingMode: none` enabled we no longer take care of finding that connection out. So, when GC happens in V8 the `implementor` instance is GC'ed. This can cause its Java counterpart instance to be collected too early and upon calling of the `callback` an "Attempt to use cleared object reference" error is very likely to occur. 
+
+### Solution
+
+To fix the previous example you must keep implementor from being GC'ed, by attaching it to the global object (or some other long-lasting object, exported module or wherever you see fit) which is not being GC'ed ever. Below, the `implementor` is attached to `global` for the sake of the example:
```
elena-p: `ever` -> `prematurely`

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+var implementor = new android.native.Implementor();     // native class
+  var callback = new android.native.NCallback({         // native interface
+    getMessage: function () {
+        implementor.getMessage();
+    }
+});
+android.native.Executor.printWithDelay(callback, 3s);
+```
+
+The implementor is enclosed by the callback implementation, but with `markingMode: none` enabled we no longer take care of finding that connection out. So, when GC happens in V8 the `implementor` instance is GC'ed. This can cause its Java counterpart instance to be collected too early and upon calling of the `callback` an "Attempt to use cleared object reference" error is very likely to occur.
```
etabakov: Clarifying who are "we" in this case...

```suggestion
The implementor is enclosed by the callback implementation, but with `markingMode: none` enabled the framework no longer takes care of finding out that connection. So, when GC happens in V8 the `implementor` instance is GC'ed. This can cause its Java counterpart instance to be collected too early and upon calling of the `callback` an "Attempt to use cleared object reference" error is very likely to occur. 
```

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+    }
+});
+android.native.Executor.printWithDelay(callback, 3s);
+```
+
+The implementor is enclosed by the callback implementation, but with `markingMode: none` enabled we no longer take care of finding that connection out. So, when GC happens in V8 the `implementor` instance is GC'ed. This can cause its Java counterpart instance to be collected too early and upon calling of the `callback` an "Attempt to use cleared object reference" error is very likely to occur. 
+
+### Solution
+
+To fix the previous example you must keep implementor from being GC'ed, by attaching it to the global object (or some other long-lasting object, exported module or wherever you see fit) which will ensure it will not be GC'ed prematurely. Below, the `implementor` is attached to `global` for the sake of the example:
```
etabakov, line 52: Using `global` to keep the instance in most cases will force the user to free the memory at some point or a memory leak will happen. Let's try to create an example where the user uses the lifespan of a parent javascript object to clear properly the memory without manually setting it to null?

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
```diff
@@ -0,0 +1,110 @@
...
+
+Naturally, you must manage the lifecycle of the implementor and release it when it is no longer needed:
+
+```javascript
+global.implementor = null;
+```
+
+## Testing an app for `markingMode: none` related issues
+
+`markingMode: none` applies to Android only, so testing for issues implies testing your app in Android Emulator or real Android device. Because bugs of this kind occur pretty randomly, there is not a single and universal way to find them. Below are some general tips on how to test/maintain an app:
```
etabakov, line 75: We should state that this article concerns Android only somewhere in the beginning to make it clear from the start.

etabakov (commented):
Great start ... Some suggestions from me:

docs/core-concepts/android-runtime/advanced-topics/marking-mode-none.md
tbozhikov: We have stated in the Introduction that this option is "added to the Android runtime", also this article is under android-runtime/advanced topics section of the docs

elena-p (approved):

PR #1534, opened 2019-02-10
Pull request for page https://docs.nativescript.org/angular/core-concepts/application-lifecycle#nativescript-application-architecture-and-lifecycle

kichloo:
I noticed an error in the page:https://docs.nativescript.org/angular/core-concepts/application-lifecycle#nativescript-application-architecture-and-lifecycle

The excerpt from the page is as follows:
"Use Application Events
NativeScript applications have the following life cycle events.

launch: This event is raised when **application launch**.
suspend: This event is raised when the application is suspended."

Actually,it can be changed to-"application is launched".
Kindly accept the PR.
Thanks.


etabakov:
This Pull Request includes much more changes than those described in the body. Please, rebase and open a new PR.

PR #1545, opened 2019-02-15
docs: improved iOS font recognition description

ycherniavskyi:
Currently, the documentation says that because iOS font recognition use **font name** to load the font, then if fonts have the same **font name** but different **file name**, then iOS can't distinguish them, and **font name** must be changed to be different.

But actually, if additional CSS font parameter, which concrete recognition, added, then the same **font name** is not a problem.
Also, CSS code snipped for Font Awesome (fontawesome-free-5.7.2-web) with this technic added.

Example Playground - https://play.nativescript.org/?template=play-tsc&id=PlfxSn&v=2

### DIFF
--- a/docs/ui/icon-fonts.md
+++ b/docs/ui/icon-fonts.md
@@ -40,7 +40,19 @@ In the above example, the `fa-brands-400.ttf` (as downloaded from the FontAwesom
 |**fa-solid-900** | Font Awesome 5 Free
 |**fa-regular-400** | Font Awesome 5 Free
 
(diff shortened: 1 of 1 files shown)


NickIliev (approved):

etabakov:

Regards from the whole team 🥇

NickIliev:
@ycherniavskyi we are reverting this PR as in the case of FontAwesome 5 there are different fonts (e.g. FontAwesome **_Brands_** and FontAwesome **_Regular_**) which are providing different sets of icons and the only solution on iOS is to manually rename the font name (as described in the original article). 

Font weight won't do the job as the icons from **_Brands_** are not present in the **_Regular_** set (and via verse)

PR #1569, opened 2019-03-01
docs(ui/styling): add extension to cssFileName

mudlabs:
There's an inconsistency in NativeScript when passing file path strings to methods. Sometimes you include the file extension, and others you don't. This change clarifies that here, the extension `.css` must be included in the file path string.


mudlabs:
@cla-bot check

NickIliev:
@mudlabs thank you for your contribution! The snippet is showing how to use the method while passing a variable called `cssFileName`

as in
```JavaScript
let cssFileName = "my-page.css";
```

That said the code **without** quotes will throw an error as you are trying to pass as an agrument an object with `css` property. Or you should add the path as a string so this is why I am closing the PR.

PR #1582, opened 2019-03-12
Revert "docs: improved iOS font recognition description"

tsonevn:
Reverts NativeScript/docs#1545


ycherniavskyi:
Sorry, do I understand it correctly that this revert pull request is canceled?

NickIliev:
@ycherniavskyi sorry - I've re-tested the described scenario and it seems that your approach is working as expected so no action is needed (you current revision will stay published in the documentation - we've canceld the revert). Again sorry for the hassle.

PR #1611, opened 2019-04-02
chore: add hmr related warning when using RadSideDrawer as root view

hettiger:
## PR Checklist

There's no warning about HMR incompatibility with RadSideDrawer as root view. 

See: https://github.com/NativeScript/NativeScript/issues/6398

Now there is a warning.

### DIFF
--- a/docs/ui/professional-ui-components/SideDrawer/show-over-navi-bar.md
+++ b/docs/ui/professional-ui-components/SideDrawer/show-over-navi-bar.md
@@ -37,7 +37,7 @@ For example:
 
 By design when navigation is executed in the NativeScript application that navigation will be executed inside the `Frame` instance which means that the {% typedoc_link classes:RadSideDrawer %} instance will be persisted during that navigation.
 
-When using the {% typedoc_link classes:RadSideDrawer %} as a root element it will by design **always be shown over the navigation bar**. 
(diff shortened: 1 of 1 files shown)


NickIliev:
@hettiger I think the warning in the documentation is not needed as we are currently working on HMR improvements including when the root is RadSideDrawer or TabView.

In fact, the HMR should work with RadSideDrawer even as we speak but with no state management and with some cases where a restart ()instead of reloading) needed. Can you please let us know what the specific issue with RadSideDrawer and HMR is and if there is a bug you can log it in nativescript-cli repository

hettiger:
@NickIliev 

> In fact, the HMR should work with RadSideDrawer even as we speak but with no state management and with some cases where a restart ()instead of reloading) needed.

HMR was not working at all but logs looked very promising and I didn't understand why I would not see any changes applied. All documents related to HMR were making me beliefe I could use it. (Out of beta, etc.) I was debugging for quite some time until I found the mentioned ticket that made me ask in Slack.

Maybe it will be working in the next release but the warning might save some people a lot of time and headaches.

Please reconsider adding it temporarily.

vchimev:
Hey @hettiger,

I opened https://github.com/NativeScript/docs/pull/1639 reflecting the state of HMR with 5.4.

I would appreciate your opinion on it.

Thanks!

hettiger:
@vchimev sorry for the late reply. LGTM but I'm not sure how this connects to my proposal.

vchimev:
In the latest version 6.0 of NativeScript, changes in an application with a side drawer as root apply at runtime. For Angular - the application bootstraps on a change.

Please, log any issues you experience as bug reports.

PR #1643, opened 2019-04-30
Explain that form fields must have values to be shown

lonerzzz:
I was hitting a missing form and trying to determine the cause because I would expect an error or a blank value if a field cannot be rendered using the provided value. This explanation will help other people who are attempting to use these forms.

- [X ] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.

The current document does not indicate that values must exist for the form and form entry to be shown so a person must guess why their form is not displayed especially if they have only created the source object and not populated it with any values.

An explanation is added so that people using the form can quickly determine why a form may not be shown.

N/A


etabakov (approved):

Regards from the whole team 🥇

PR #1656, opened 2019-05-15
Update extend-application-activity.md

shiv19:
the old method signature was deprecated with NS 5.4
```AndroidActivityCallbacks.onCreate(activity: any, savedInstanceState: any, superFunc: Function) is deprecated. Use AndroidActivityCallbacks.onCreate(activity: any, savedInstanceState: any, intent: any, superFunc: Function) instead.```

## PR Checklist

### DIFF
--- a/docs/core-concepts/android-runtime/advanced-topics/extend-application-activity.md
+++ b/docs/core-concepts/android-runtime/advanced-topics/extend-application-activity.md
@@ -117,7 +117,7 @@ The core modules ship with a default `android.support.v7.app.AppCompatActivity` 
                 frame.setActivityCallbacks(this);
             }
             // Modules will take care of calling super.onCreate, do not call it here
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by NickIliev — COMMENTED** [review]
Hey, @shiv19 thank you for the PR - indeed the overload without intent param is now deprecated, and we should use the method with the intent param.

Would you mind if you also update the TypeScript code in the same PR
It should be called as done [here](https://github.com/NativeScript/NativeScript/blob/3efc06ed986b114e7ff52764e3cb78c07ad7c8c9/tns-core-modules/ui/frame/activity.android.ts#L29)

```TypeScript
this._callbacks.onCreate(this, savedInstanceState, this.getIntent(), super.onCreate);
```

shiv19:
@NickIliev Thanks!
I've updated the TS code :)

NickIliev (approved):

PR #1760, opened 2019-09-02
docs: hide the Enterprise Auth articles from Sidekick docs

ggarabedian:


ggarabedian:
@tsonevn or @NickIliev, can you please check if the redirects are done correctly?

PR #1763, opened 2019-09-04
fix: redirect a couple of Angular-specific articles to the Angular flavour of the docs

sis0k0:
This PR changes the NGINX configuration to redirect:

- https://docs.nativescript.org/performance-optimizations/lazy-loading to https://docs.nativescript.org/angular/performance-optimizations/lazy-loading
- https://docs.nativescript.org/core-concepts/angular-navigation to https://docs.nativescript.org/angular/core-concepts/angular-navigation

### DIFF
--- a/build/nginx.conf
+++ b/build/nginx.conf
@@ -118,6 +118,12 @@ http {
 		# Redirect Code Sharing story to Angular flavour
 		rewrite ^/code-sharing/intro $proto://$host_and_port/angular/code-sharing/intro permanent;
 
+        # Redirect Lazy Loading article to Angular flavour
+		rewrite ^/performance-optimizations/lazy-loading $proto://$host_and_port/angular/performance-optimizations/lazy-loading permanent;
+
+        # Redirect Angular Navigation article to Angular flavour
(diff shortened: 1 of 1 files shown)


etabakov:
Why are these redirects needed?

sis0k0:
The articles are Angular-specific and are in the Angular flavour of the docs. Also, the Google results currently lead to the non-Angular URL and you end up in the NativeScript Core docs instead of NativeScript Angular. Third, you don't see the article in the side navigation if you open the non-Angular URL:

69427304_2552812338109874_1190712350613700608_n

etabakov:
The real fix for that is to modify the Jekyll build to exclude articles that are not for the respective flavour from the output. Adding redirects will only partially treat the symptoms, but won't solve the problem.

sis0k0:
@etabakov, that sounds better indeed. How can we do that?

etabakov:
One idea that I have is to modify the FlavorGenerator plugin to delete the page if it doesn't matches the current flavor: https://github.com/NativeScript/docs/blob/master/build/_plugins/flavor_generator.rb

tsonevn (approved):

PR #1787, opened 2019-09-24
feat: highlightColor info

tsonevn:
adding highlightColor info


MartoYankov:
See https://github.com/NativeScript/nativescript-sdk-examples-ng/pull/416 for comments

PR #1800, opened 2019-09-30
update screen size page

rdlauer:
commenting out the notes about the incompatibility with webpack


NickIliev:
Hey, @rdlauer  I've just seen this PR (after opening [this one](https://github.com/NativeScript/docs/pull/1801)) - the thing is that those qualifiers are still not supported by Angular (but now supported by Core + Webpack) - so the notes should be present but updated.

Closing in favor of https://github.com/NativeScript/docs/pull/1801

Issue #1182, opened 2018-06-08
Feedback form 06.2018


NickIliev:
- [x] **Fix** - Broken links https://docs.nativescript.org/ui/layouts#layout-pass

- [x] **Improve** - add [import declaration](https://docs.nativescript.org/angular/code-samples/ui/image) @tsonevn 

- [ ] **Improve** - add CLI install instructions / link https://docs.nativescript.org/sidekick/intro/requirements @ggarabedian 

- [ ] **Improve** - update CLI required version https://docs.nativescript.org/sidekick/intro/requirements @ggarabedian 

- [x] **Improve** - add an explanation about the generated link that should be manually entered in Chrome https://docs.nativescript.org/tooling/chrome-devtools @tsonevn 

- [x] **Improve** - include `android` import comment (tns-platform-declaration comment in the snippet) https://docs.nativescript.org/ui/placeholder @tsonevn 

- [x] **Improve** - add `PUT` and `DELETE` examples if possible to use https://docs.nativescript.org/angular/code-samples/http @NickIliev

- [x] **Improve** - 'ui/build' section (add detailed example e.g. how to add view-model or event listener) https://stackoverflow.com/a/39844818/4936697 @NickIliev resolved via [https://github.com/NativeScript/docs/pull/1323](https://github.com/NativeScript/docs/pull/1323)

PR #1826, opened 2019-10-15
docs: Added CLI Version and CLI Installation reference

shivamkainth:
[#1182 Added CLI Version}
Went to npmjs website, confirmed the latest version from there and then updated in the docs.
URL: https://www.npmjs.com/package/nativescript
Also added the CLI installation reference link.
#1182

## PR Checklist

### DIFF
--- a/docs/sidekick/intro/requirements.md
+++ b/docs/sidekick/intro/requirements.md
@@ -31,8 +31,9 @@ slug: requirements
 <p></p>
 
 ## Additional Requirements
-
+### Install Instructions:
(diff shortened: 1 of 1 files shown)


shivamkainth:
@cla-bot check

tsonevn:
HI @shivamkainth,
Thank you for the provided PR. However, I do not think that referring to a specific CLI version inside the article is the best practice. Perhaps, setting an NPM link to the nativescript package would be a better solution. Then every docs visitor will be able to check which version is the latest.

tsonevn (changes requested):
Set an NPM link to the nativescript package, instead of a specific version

NickIliev (approved):

PR #1881, opened 2020-03-28
feat: No need to show 'Combining Nested Navigation Scenarios' section in Angular guide

s-patompong:
Since there is no Angular code under 'Combining Nested Navigation Scenarios' section, I don't think we should show it under angular doc.

It confused me when I see only header like this without the body:

Screen Shot 2563-03-28 at 19 21 14

## PR Checklist

- [X] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [X] If there is an issue related with this PR, point it out here.

I see a section called 'Combining Nested Navigation Scenarios' in [Angular guide](https://docs.nativescript.org/angular/core-concepts/nested-navigation#nesting-lateral-in-lateral), however I don't see the section's body (describing code, etc.). However, both title and body are shown under normal [NativeScript guide](https://docs.nativescript.org/core-concepts/nested-navigation#nesting-lateral-in-lateral).

I think it'd make more sense to bring back this section in Angular guide once the code is ready. For now we better hide it so there is no confusion. Otherwise, we can show it but at least show some text saying things like 'For this section, only NativeScript is available, click here to see it', or something like that.

I make the webpage hide the 'Combining Nested Navigation Scenarios' section in the Angular guide, it'll still show up as usual under normal NativeScript guide.

### DIFF
--- a/docs/core-concepts/nested-navigation.md
+++ b/docs/core-concepts/nested-navigation.md
@@ -224,10 +224,10 @@ TabView (lateral navigation)
 [Playground Demo Angular](https://play.nativescript.org/?template=play-ng&id=ObeDAp)
 {% endangular %}
 
-
+{% nativescript %}
 ## Combining Nested Navigation Scenarios
 
-{% nativescript %}
+
 ![navigation-schema-backward](../img/navigation-extended/navigation-examples-page-7.png?raw=true)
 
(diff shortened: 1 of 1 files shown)


s-patompong:
@cla-bot check

NickIliev (changes requested):
If you want to show something explicitly for NaitveScript core while using the markdown injector then you need to have opening ( like **_{% nativescript %}_** ) **and** closing tag ( like **_{% endnativescript %}_** )

s-patompong:
Line 259 already has `{% endnativescript %}`, my code is basically move `{% nativescript %}` up one line so instead of showing only the header of **Combining Nested Navigation Scenarios** section, it just hide it totally so user don't get confused when reading Angular doc.

PR #1884, opened 2020-04-19
update-quicksetup-md

ctzn1984:
I ran into the issue as described and felt it would be a good idea to include this note for those users who run into the error I've seen.  Thanks in advance! Love the framework by the way! :)

## PR Checklist

- [X] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.

I noticed during my install I received an error related to the '@powershell' command.

I have added a note under the command for installing the full setup for windows in powershell which will help some users correct the error if they encounter it on their systems.


NathanaelA:
@ctzn1984 - Thanks!

PR #1946, opened 2021-02-21
fix(docs): remove cask from brew commands

fadarrizz:
It is no longer required to install cask packages with `brew cask`, since it has been integrated within brew itself in version 2.5.

See the release notes for more information
https://brew.sh/2020/09/08/homebrew-2.5.0/

## PR Checklist

Many brew cask commands have been integrated with brew. The commands that use cask in the documentation don't work anymore.

The two commands that make use of cask have been updated, so that they work again.

### DIFF
--- a/docs/start/ns-setup-os-x.md
+++ b/docs/start/ns-setup-os-x.md
@@ -102,7 +102,7 @@ Complete the following steps to setup NativeScript on your macOS development mac
     <pre class="add-copy-button"><code class="language-terminal">brew tap AdoptOpenJDK/openjdk
     </code></pre>
 
-    <pre class="add-copy-button"><code class="language-terminal">brew cask install adoptopenjdk8
+    <pre class="add-copy-button"><code class="language-terminal">brew install adoptopenjdk8
     </code></pre>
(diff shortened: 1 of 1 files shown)


rigor789:
I believe it has been deprecated, and replaced by the `--cask` flag, so it should be

```bash
brew install --cask adoptopenjdk8
```

instead? (https://formulae.brew.sh/cask/adoptopenjdk#default)

image

fadarrizz:
> I believe it has been deprecated, and replaced by the `--cask` flag, so it should be
> 
> ```shell
> brew install --cask adoptopenjdk8
> ```
> 
> instead? (https://formulae.brew.sh/cask/adoptopenjdk#default)
> 
> image

Interesting, I ran both commands without the --cask flag and that seems to work as well.

But I believe it's better to follow the provided installation guidelines from brew itself, so I shall add the flag.

fadarrizz:
@cla-bot check

rigor789 (approved):

rigor789:
Thanks @Fadarrizz!

PR #1957, opened 2021-06-14
Javascript examples fixed

maamarcos:
## PR Checklist


rigor789:
The `import` statements are valid es6 code - so all the example should work as-is - `require` is supported, but is considered "legacy", and es6 is preferred.

